PR #40, opened 2026-03-06
feat(gcs): enhance signed URL generation for WIF support

pedrojreis:
Since our last contribution some things changed, packages where updated and dotenv got added.
The latest image 3.3 broke WIF and this will fix it as we tested in our k8s cluster environment

This pull request updates the logic for generating signed URLs for Google Cloud Storage (GCS) uploads to better support environments using Workload Identity Federation (WIF) or other federated credentials. The key improvement is enabling signed URL generation even when a local private key is not available, by delegating signing to the IAM signBlob API as needed.

**GCS signed URL generation improvements:**

* Updated `upload_to_gcs` in `gcs.py` to detect when federated credentials (such as WIF) are used, and to generate signed URLs via the IAM signBlob API by providing the service account email and access token, ensuring compatibility with both key-file and federated authentication methods.

### DIFF
--- a/upload_tools/backends/gcs.py
+++ b/upload_tools/backends/gcs.py
@@ -38,12 +38,31 @@ def upload_to_gcs(file_object, file_name: str, gcscfg, signed_url_expires_in: in
         file_object.seek(0)  # Reset file pointer to beginning
         blob.upload_from_file(file_object, content_type=content_type)
 
-        # Generate a signed URL valid for configured duration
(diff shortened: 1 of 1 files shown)


dvejsada:
@pedrojreis Could you please just have a look if the comments are relevant or not? Thanks!

pedrojreis:
@dvejsada Did not accept as they broke on our cluster.

PR #51, opened 2026-03-24
feat(upload): add optional filename parameter for file uploads

yanivvds:
## Summary
This PR adds an optional `file_name` parameter to all document creation tools and dynamic template tools.

If a name is provided, uploaded files will use a more readable format like `a1b2c3d4_Q1_Sales_Report.docx` instead of a long UUID-only filename. If no `file_name` is passed, the current behavior stays exactly the same.

## Why
Right now, uploaded files are stored with UUID-only names, which makes them hard to recognize when browsing storage or downloading them later.

This change makes it possible to keep filenames meaningful and easier to identify, while still preserving uniqueness.

## What changed
- Added filename sanitization and named object generation helpers
- Added an optional filename parameter to `upload_file()`
- Added `file_name` support to all static MCP document tools
- Passed `file_name` through the DOCX, PPTX, XLSX, XML, and email upload flows
- Updated dynamic DOCX tools to use `file_name`, or fall back to the template name
- Updated dynamic email tools to use `file_name`, then subject, then template name

## Filename handling
Provided filenames are sanitized before upload:
- unsafe characters are removed
- spaces are converted to underscores
- names are truncated to 100 characters
- empty results fall back to `"document"`
- an 8-character UUID prefix is added to keep names unique

## Test plan
- [ ] Create a Word document with `file_name` and verify the uploaded blob includes it
- [ ] Create a Word document without `file_name` and verify UUID naming still works
- [ ] Test special characters in `file_name` (for example: `"Report (Q1) — Final!"`) and verify sanitization
- [ ] Test each document type: DOCX, XLSX, PPTX, EML, and XML
- [ ] Test dynamic DOCX and email template tools and verify fallback naming works correctly

### DIFF
--- a/docx_tools/base_docx_tool.py
+++ b/docx_tools/base_docx_tool.py
@@ -25,7 +25,7 @@ logger = logging.getLogger(__name__)
 
 
 def markdown_to_word(markdown_content, title=None, author=None, subject=None,
-                     header_text=None, footer_text=None, include_toc=False):
+                     header_text=None, footer_text=None, include_toc=False, file_name=None):
     """Convert Markdown to Word document."""
     logger.info("Starting markdown_to_word conversion")
     path = load_templates()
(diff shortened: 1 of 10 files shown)


dvejsada:
Thanks for the PR! I will review over the weekend.

PR #65, opened 2026-06-17
feat(xlsx): pure-Python formula recalculation, financial-modeling mode, and circular-reference detection

andormarkus:
## Summary

Adds a substantial upgrade to the Excel (`create_excel_from_markdown`) tool: in-process formula **recalculation** with cached-value injection, a **financial-modeling mode** with CFA-standard color coding and number formats, and independent **circular-reference detection** — plus a set of correctness fixes to the formula-reference and `types`-directive pipelines. No external binary is required; recalculation uses the pure-Python [`formulas`](https://github.com/vinci1it2000/formulas) library.

### Formula recalculation (cached values, pure-Python)
- New `xlsx_tools/formula_engine.py` evaluates every formula in-process and `xlsx_tools/xml_cache.py` injects the computed values back into the worksheet XML (`<v>` tags) so files **preview correctly** in non-recalculating clients (Google Sheets preview, mail clients, `openpyxl` with `data_only=True`).
- Caches numeric, boolean, datetime, **and** string results (string formulas use OOXML inline `t="str"` cells).
- Runs in a worker thread bounded by a configurable timeout; on timeout the file ships without cached values (the spreadsheet application recalculates on open).
- **Error policy:** when `recalc=True` is explicitly passed and errors are detected (`#REF!`, `#DIV/0!`, `#VALUE!`, `#NAME?`, `#NULL!`, `#NUM!`, `#N/A`, plus `#CIRC!`), the call fails with a type-grouped summary so the model can fix and retry (zero-errors standard). When recalc runs as a default, errors are logged but the file is still delivered.
- External-workbook links (`=[Other.xlsx]Sheet1!A1`) are isolated so a single unsupported reference can't suppress caching for the whole workbook (they are evaluated when the file is opened).

### Circular-reference detection (independent of the engine)
- DFS graph analysis over the formula strings catches cycles that the engine silently resolves to nothing. Runs **even when recalc is disabled**. Robust reference parsing (string-literal stripping, coordinate validation, range/3D/absolute refs) avoids false positives.

### Financial-modeling mode (`financial_modeling=true`)
- CFA-standard color coding: blue inputs, black local formulas, green cross-sheet links, **red external-workbook links**, yellow background for source-cited cells.
- Number-format conventions: `$#,##0` currency (with `integer` and `dash`/`parens` variants), `0.0%` percentages, `0.0x` multiples, zeros-as-`-`, negatives in parentheses. 4-digit-year strings treated as text labels.
- `sources:` directive attaches source-citation cell comments.

### Directives & parsing improvements (`helpers.py`, `parser.py`)
- `types:` directive supports `text, currency:<sym>[:variant], number[:fmt|variant], percent[:variant], date[:fmt], bool, multiple`.
- Fixes in this branch:
  - Formula cells in **typed columns** now resolve relative/table refs (`B[0]`, `T1.B[0]`, `Sheet!T1.B[0]`) and apply the column's number format (previously written verbatim → `#NAME?`).
  - `types` directive no longer splits on commas **inside** a number format (`number:#,##0`), which had truncated formats and shifted columns (silent data loss).
  - `currency:$:integer:dash` now applies the dash/parens sections alongside the integer precision (was dropped).
  - Typed numeric columns honor the financial dash default under `financial_modeling`, matching the untyped path.
  - Cross-sheet `SUM` range prefix, current-row vs table-relative ref resolution, sheet-name collision warnings, unresolved-reference warnings.

### Config / packaging
- New env vars: `XLSX_RECALC_ENABLED` (default `true`), `XLSX_RECALC_TIMEOUT_SECONDS` (default `30`), `XLSX_DEFAULT_FONT`.
- Adds `formulas>=1.3.4` to `requirements.txt`.
- Dockerfile switches `python:3.12-alpine` → `python:3.12-slim` because `formulas` pulls in numpy/scipy, which ship pre-built wheels only for glibc (avoids ~400MB of Alpine build deps); user creation updated for Debian (`groupadd`/`useradd`).

## Test plan
- [x] `pytest` xlsx suite — **291 passing** across `test_xlsx_creation.py`, `test_xlsx_financial.py`, `test_xlsx_recalc.py`, `test_xlsx_edge_cases.py` (new + expanded coverage: recalc happy-path/errors/timeout, circular detection + no-false-positives, cached string/numeric/datetime injection, financial color coding + formats, typed-column formula resolution, comma-in-format parsing).
- [x] Verified cached values render via `openpyxl` `data_only=True`.
- [x] Confirmed no regressions elsewhere (only pre-existing, unrelated `test_run_blocking.py` email-template failures remain).
- [ ] Reviewer: build the `-slim` Docker image and confirm `formulas`/numpy/scipy install from wheels.
- [ ] Reviewer: smoke-test a financial model end-to-end (color coding + cached values) in Excel and Google Sheets preview.


dvejsada:
@andormarkus Thank you for the PR. Here are my first comments:
1. Formula recalculation - I like this, but I would prefer this to be set only via ENV variables and not included in tool arguments.
2. Circular-reference detection - OK
3. Financial-modeling mode - This is my main pain point. I do not want the tool to be aligned to any specific usage. This should be achieved on client side. I would prefer to keep the server usage agnostic and only give the MCP server general tools to achieve the desired outcome - e.g. ability to set background or font color to a cell, or set custom excel template with styles.
4. Directives & parsing improvements - OK

I would suggest to remove 3 from the PR at this time, pending further discussion on how to achieve this. After 3 is removed and 1 is updated as suggested, I will be happy to merge (pending testing).

andormarkus:
Hi @dvejsada, thank you for the clear feedback!

I've addressed all your points:

1. **Formula recalculation** — the `recalc` tool argument is removed. Recalculation is now controlled entirely via `XLSX_RECALC_ENABLED` (existing) and a new `XLSX_RECALC_STRICT` env var (default `false`). When strict mode is off, errors are logged but the file is always delivered; when strict is on, errors fail the call — preserving the zero-errors option without any tool-level knob.

2. **Circular-reference detection** — unchanged, as agreed.

3. **Financial-modeling mode** — fully removed. `apply_financial_styling`, all CFA color constants, the years-as-text rule, and the `financial_modeling` parameter are gone from both the tool API and `helpers.py`. General-purpose capabilities (number-format variants, `multiple`/`12.5x` type, `sources:` cell comments, default font) are kept since they're usage-agnostic. The separate discussion on cell-level background/font color and custom templates can happen in a follow-up PR.

4. **Directives & parsing** — unchanged, as agreed.

Test coverage is maintained — `test_xlsx_financial.py` is replaced by `test_xlsx_styling.py` (general tests only), and the recalc tests are rewritten to drive behavior via config patching. All 554 tests pass.

Happy to address any further comments!

dvejsada:
@andormarkus Thank you very much for quick turn around. I will test that everything works as expected and will merge (presumably over the weekend).

dvejsada:
@claude please review this PR for correctness bugs, edge cases, security issues, and clear simplifications. Post findings as inline review comments.

dvejsada:
@andormarkus Please kindly have a look on PR comments if they need to be addressed. Thanks!

andormarkus:
Hi @dvejsada, sorry for the late update — I was out of weekly tokens.

I've addressed the findings from the Claude review (836b8fe):

**Critical**
- **Timeout now actually works.** The `with ThreadPoolExecutor` form was calling `shutdown(wait=True)` on `__exit__`, which blocked past the timeout. Rewritten to manage the executor manually.
- **No more RecursionError on large workbooks.** Converted the circular-ref DFS from recursive to iterative (verified with a 2000-cell chain that would have hit Python's 1000-frame limit).

**Medium**
- **Error detection fixed.** `_is_error_string` was using a substring match that false-flagged legitimate strings like `"#N/A not found"`. The exact-match fix needed `str(value)` because the `formulas` library returns errors as a custom `XlError` str-subclass with overridden `__eq__`.
- **No more caller-file corruption.** Path inputs to `recalculate_workbook` are now always copied to a temp file before external-link blanking.
- **No more silent sheet-wide drops.** `_format_float` now guards `nan`/`inf` with `math.isfinite` (previously raised and dropped all cached values for the worksheet).
- **XML parsing hardened.** `xml_cache` now uses `defusedxml.ElementTree` for parsing (already a dependency).

**Low** — duplicate import, step-numbering, and comment cleanups.

All 264 xlsx tests pass (625 total; only the pre-existing unrelated `test_run_blocking.py` email failures remain).

dvejsada:
@claude review

andormarkus:
Hi @dvejsada — followed up on the nan/inf finding from the latest Claude review (1f83c77).

**The issue with my previous fix:** the `math.isfinite` guard in `_format_float` stopped the crash, but the result was still written as `<v>nan</v>` into a numeric cell. Per ECMA-376, `xsd:double` special values are case-sensitive (`NaN` / `INF` / `-INF`), so Python's lowercase `repr()` output is non-conformant and could make Excel flag the file for repair on open — worse than the original bug.

**The fix:** non-finite floats are now caught in `formula_engine.recalculate_workbook` right after `_unwrap_scalar()`, alongside the existing string-error check, and recorded as `#NUM!` errors so they never reach `xml_cache`. This matches how `XlError` sentinels are already handled — the cell is treated as a formula error, and Excel recalcs it natively on open to display `#NUM!` as expected.

The `_format_float` guard stays as defensive-only (in case the function is ever called directly), with an updated docstring noting the engine layer is the primary handler.

Added two tests covering the path. All 266 xlsx tests pass.

This should close out all the findings from both review rounds — happy to do another pass if anything else comes up.

dvejsada:
@claude review

dvejsada:
@andormarkus could you please have a look on the second round of comments so we can finalize and merge it?

dvejsada:
Mostly implemented by #88 and #89, only outstanding item is formula calculation in output - the proposed implementation is compute heavy and has issues with license in dependencies. Please kindly file an implementation plan in issues first so we can confirm before new PR.

PR #70, opened 2026-06-19
fix(docker): patch base-image OS CVEs (openssl/musl/xz/sqlite)

dvejsada:
## Summary
Patches the 18 image-scan CVEs, **all of which are Alpine OS packages** in the pinned base (`python:3.12.8-alpine`) — none are in app code or Python dependencies.

- Roll the base from the stale `python:3.12.8` to `python:3.12` (latest 3.12 patch on current Alpine).
- `apk upgrade --no-cache` in both stages; the **runtime-stage** upgrade patches the shipped image (only `/opt/venv` is copied from the builder).

## Verified (local build)
Alpine `3.21` → `3.24.1`; `apk list --upgradable` is empty.

| Package | Was | Now |
|---------|-----|-----|
| openssl | 3.3.2-r4 | **3.5.7-r0** |
| musl | 1.2.5-r8 | **1.2.6-r2** |
| xz | 5.6.3-r0 | **5.8.3-r0** |
| sqlite | 3.47.1-r0 | **3.53.2-r0** |

## To ship
The Docker image only rebuilds on a **release** event, so after merge cut a patch release (e.g. `v3.9.1`) — or re-run the build workflow — to publish the patched image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,12 +4,17 @@
 # If you need more help, visit the Dockerfile reference guide at
 # https://docs.docker.com/go/dockerfile-reference/
 
-ARG PYTHON_VERSION=3.12.8
+# Roll with the latest 3.12 patch so the base ships current Alpine packages.
+# Pin by digest here if you need fully reproducible builds.
+ARG PYTHON_VERSION=3.12
 
 # =============================================================================
 # Stage 1: Builder - Install dependencies
 # =============================================================================
 FROM python:${PYTHON_VERSION}-alpine AS builder
+
+# Patch base OS packages (openssl, musl, xz, sqlite, …) to the latest available.
+RUN apk upgrade --no-cache
 
 # Prevents Python from writing pyc files.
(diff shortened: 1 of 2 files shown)


dvejsada:
Added a dependency bump to this PR (validated separately):

**`fastmcp` 3.2.0 → 3.4.2** (brings `mcp` 1.28.0) — the only genuinely outdated *pinned* direct dependency. `python-pptx` (1.0.2) and `python-docx` (1.2.0) are already at latest, and the `>=` deps already float to the newest at build time, so no other `requirements.txt` changes are needed.

Validated in a clean venv with the bumped requirements: **489 tests pass** (incl. the `run_blocking` and dynamic tool-registration suites where a framework bump would most likely break).

Note: a `pip-audit` of the local dev venv flagged some transitive-dep CVEs (cryptography, azure-core, authlib), but those are stale *dev-venv* versions — the image build pulls the patched latest for those `>=`/transitive deps. They are not introduced or fixed by this PR.

PR #71, opened 2026-06-20
feat(docx): conditional template blocks via {{#if}}/{{/if}} markers

dvejsada:
## Summary

Adds **Phase 1 block-level conditional inclusion** for dynamic DOCX templates. A boolean tool argument can now keep or drop a range of body content wrapped between marker paragraphs:

```
{{#if flag}} ... {{/if}}    keep when flag is truthy
{{^if flag}} ... {{/if}}    keep when flag is falsy
```

Everything between the markers — paragraphs, lists, page breaks and whole tables (they are sibling block elements in the body) — is pruned together, before the existing placeholder substitution runs.

## What changed

- **`docx_tools/conditionals.py`** (new) — `resolve_conditionals()` prunes body block elements using a keep/drop stack. Supports nesting and markers that Word has split across runs.
- **`docx_tools/dynamic_docx_tools.py`** — wires `resolve_conditionals()` into `_sync_impl`, before substitution (it must run first since it deletes whole block elements). `bool` args already existed, so no arg plumbing was needed.
- **`config/docx_templates.yaml`** — documents the syntax/limits and adds a commented `bool` arg example on `formal_letter`.
- **`tests/test_docx_conditionals.py`** (new) — 16 tests.

## Behaviour notes

- **Forgiving errors:** unbalanced markers log a warning, keep all content, and strip only the marker paragraphs; an unknown condition name warns and keeps the content (never silently drops).
- **Repeated markers/placeholders** behave predictably: each `{{#if flag}}` block is evaluated independently against the same value; markers pair by nesting order (a `{{/if}}` closes the most recent open).
- **Out of scope for v1** (forward-compatible with this syntax): inline spans within a paragraph, table-cell / header / footer conditionals, and boolean/comparison expressions.

## Testing

- New suite: **16 passed**.
- Full docx suite: **267 passed, no regressions**.

### DIFF
--- a/config/docx_templates.yaml
+++ b/config/docx_templates.yaml
@@ -24,11 +24,15 @@
 #
 # Notes:
 #   - The condition name must match a template argument (declare it as type: bool).
-#   - Blocks may be nested. Markers split across runs by Word are still detected.
+#     An unknown name keeps its content (so a typo never silently deletes a block).
+#   - The marker must be the paragraph's only text; spacing inside the braces is
+#     tolerated (e.g. "{{ #if flag }}"). Markers split across runs by Word are
+#     still detected. Blocks may be nested.
 #   - Granularity is block-level (whole paragraphs/tables between the markers).
 #     Conditionals inside table cells, headers and footers are not handled yet.
 #   - Unbalanced markers do not fail the document: a warning is logged, all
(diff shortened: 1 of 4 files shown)


dvejsada:
@claude review

dvejsada:
@claude review

PR #72, opened 2026-06-21
fix(docx): disambiguate numbered-list dates and preserve placeholder formatting

dvejsada:
## Summary

Two independent DOCX fixes discovered while reviewing list rendering and template behaviour.

### 1. Numbered-list vs. date disambiguation
A standalone numbered line such as a Czech date `23. června 2026` was misread as a one-item ordered list. Detection is now **continuation-aware**: a numbered line starts a list only when it begins at `1` **or** has a continuation (a sibling/nested list item).

| Input | Result |
|---|---|
| `23. června 2026` (lone) | paragraph ✓ |
| `5. Fifth` / `6. Sixth` | list, start 5 preserved ✓ |
| `1. Only item` (lone) | single-item list ✓ |
| `1. června 2026` (day-1, lone) | indistinguishable from a one-item list → escape as `1\. ...` |

The same `ordered_list_is_genuine` check is applied in both the block dispatcher and `contains_block_markdown`, so template placeholder content behaves consistently. Day-1 dates already render correctly when escaped (`1\.`) via the existing backslash-escape path.

### 2. Placeholder replacement preserves formatting
`_replace_placeholder_in_paragraph` previously flattened the surrounding text to plain runs and only carried over the placeholder's font name/size/colour. Now:
- **Surrounding inline text** (before/after the placeholder, in the same paragraph) keeps its run formatting — sliced at the placeholder boundaries with each slice's `<w:rPr>` deep-copied.
- The **placeholder run's bold/italic/underline/highlight** (alongside font/size/colour) is captured and applied to the replacement text, **only where the markdown value left it unset** — so `**bold**` in the value still wins and combines with, rather than is clobbered by, the placeholder's format.

Scope: applies to the inline replacement path; block content (list/heading values) is still inserted as separate paragraphs styled by `style_map`, as before.

## Tests
- `tests/test_docx_ordered_list_date.py` — 12 cases (helper + full pipeline): lone dates, start-at-5 lists, single-item lists, escaped day-1 date, documented two-adjacent-dates limitation.
- `tests/test_docx_placeholder_formatting.py` — 8 cases: surrounding bold/italic/colour preserved, placeholder bold/italic/underline/highlight/font/size applied, markdown-vs-placeholder combination, split-placeholder.
- Full suite: **533 passed**.

## Docs
README updated for both behaviours (numbered-list/date rule + escape, and placeholder formatting preservation).

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -242,6 +242,8 @@ Both the Word and Excel tools accept Markdown. These references cover **everythi
 | `# H1` … `###### H6` | Headings 1–6 |
 | `- item` / `* item` / `+ item` | Bullet list (nest by indenting children — 2-4 spaces or a tab → `List Bullet 2/3`) |
 | `1. item` / `2. item` | Numbered list (nest by indenting children). **Numbering restarts automatically** whenever a list begins again with `1.` |
+
+> 💡 A single numbered line is only treated as a list when it starts at `1.` **or** is followed by another item. This means a standalone date like `23. června 2026` renders as plain text, not a list. The one exception is a day-1 date (`1. června 2026`), which is indistinguishable from a one-item list — escape the dot to keep it as text: `1\. června 2026`.
 | `> quote` | Block quote (`Quote` style) |
 | `\| A \| B \|` + `\|---\|---\|` | Table (see table features below) |
 | ` ``` ` … ` ``` ` (or `~~~`) | Fenced code block — content is rendered verbatim in a monospace font and **not** parsed as markdown |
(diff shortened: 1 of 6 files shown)


dvejsada:
@claude review

dvejsada:
Thanks for the review — addressed in 4048d7e:

- **#1 README table break (bug):** moved the numbered-list/date tip out of the "Block elements" table to a standalone note after the `***` row, alongside the existing `⚠️` tip, so the table no longer gets cut in half.
- **#3 / #7 explicit `bold=False` not propagated:** `_apply_placeholder_format` now propagates the placeholder's explicit emphasis value for bold/italic/underline (using `is not None` instead of a truthy check), so an explicit "no bold" passes through to plain replacement text — while markdown formatting in the value still always wins. Added two tests: explicit-not-bold propagation, and markdown-bold-wins-over-not-bold. (Highlight kept as a truthy check to avoid propagating `AUTO`.)

Points #2, #4, #5, #6 were confirmations with no change requested. Full suite: 535 passed.

dvejsada:
@claude review

PR #75, opened 2026-06-23
fix(docx): correct literal \n handling and unify newline semantics

dvejsada:
## Problem

A literal `\n` (backslash + `n` — e.g. from a double-escaped tool argument) was collapsing to a stray `n` and the paragraph break was lost. Root cause: the backslash-escape handler (`_ESCAPE_RE = r'\(.)'`) stripped the slash before **any** character (so it also corrupted `\t`, Windows paths like `C:\new`, etc.). The literal `\n` itself was largely a side effect of tool/argument descriptions demonstrating `\n` as if it were syntax, which taught models to emit it.

On top of that, newline handling **diverged** between the classic tool and templates: in a template, the same prose rendered as one paragraph with `<w:br>` (inline path) or as separate paragraphs (when the value happened to also contain a list), while the base tool always treated a line break as a new paragraph.

## Changes

**Parser correctness**
- Restrict `_ESCAPE_RE` to ASCII punctuation (CommonMark), so a backslash before a letter/digit is preserved instead of silently dropped.
- Add `normalize_escaped_newlines()` converting literal `\n` / `\r\n` / `\r` to real newlines; applied at all content entry points (inline, block processor, template placeholder routing).

**Unified newline semantics (base tool == templates)**
- A multi-line whole-paragraph placeholder now renders through the same `process_markdown_content` pipeline as the base tool: **line break = new paragraph**, **`<br>` / two trailing spaces = soft break**, **blank line = new paragraph**.
- The placeholder paragraph's style and run format are propagated to plain (Normal) paragraphs, while headings/lists/quotes keep their own styles.
- Single-line values and mid-paragraph placeholders (e.g. `Dear {{name}},`) stay inline (soft breaks only).
- Removed the now-redundant inline paragraph-split helper.

**Docs / prompts**
- `create_word_from_markdown` description no longer demonstrates a literal `\n`.
- `config/docx_templates.yaml`: documents the three break cases and the YAML `|` vs `>`/`>-` scalar trap; adds a `letterhead_breaks_demo` template.

## Side benefit
Body paragraphs are now separate `<w:p>` instead of one paragraph with `<w:br>`, so justified ("do bloku") text no longer stretches the line before each break.

## Tests
223 passing, incl. new `tests/test_docx_escaped_newlines.py` (literal `\n` no longer corrupts; escaping of punctuation still works; base↔template parity; style propagation vs. heading style retention).


dvejsada:
@claude review

dvejsada:
Thanks for the review — addressed in bfe7633:

- **`_propagate_format_to_block` — placeholder's named style leaking onto prose (main item):** Fixed. Now only the *layout* (`alignment`/`indent`/`spacing`) is copied from the placeholder paragraph; its `<w:pStyle>` is dropped from the deep-copy before insertion, so a placeholder sitting in a `Heading`/named-style paragraph no longer stamps that style onto produced prose. Produced prose keeps its own default style. Added `test_template_named_style_not_stamped_on_prose`.
- **`_ESCAPE_RE` regex opacity:** Added a comment explaining `re.escape` is used so `]`, `\`, `^`, `-` are literal inside the `[...]` class.
- **Double normalization:** Clarified with a comment — the early `normalize_escaped_newlines()` in `_replace_placeholder_in_paragraph` is needed for *routing* (block detection + the multi-line check); the downstream calls in `process_markdown_content`/`parse_inline_formatting` are what the base tool relies on and are idempotent here.
- **`letterhead_breaks_demo`:** Marked as documentation-only in its `description`/title.
- **Test gap — literal `\r\n` through a template placeholder:** Added `test_template_placeholder_literal_crlf`.
- **Block content in mid-paragraph position:** Acknowledged as pre-existing behaviour (unchanged), as noted.

Full suite: 225 passing.

dvejsada:
@claude review

dvejsada:
Second review addressed in f9da538:

- **`_propagate_format_to_block` — strip `<w:numPr>` too (main item):** Done. The layout copy now drops both `<w:pStyle>` and `<w:numPr>`, so a placeholder that sits in a list item no longer stamps list numbering onto produced prose. Added `test_template_list_numbering_not_stamped_on_prose` (constructs a placeholder paragraph with direct numbering and asserts produced prose has none).
- **`prilohy` default `" "` nit:** Added an inline comment explaining a space keeps the placeholder from rendering as-is when omitted.
- **`test_literal_newline_renders_as_break` clarification:** Confirmed against the python-docx version in use (1.2.0): a `\n` in `run.text` becomes a `<w:br>` and is reflected back as `\n` in `Paragraph.text`, so both assertions hold together. Added a comment in the test noting this.

Full suite: 226 passing.

PR #76, opened 2026-06-24
fix(docx): recognise lists after <br> and continue numbering across headings

dvejsada:
## Problem

Two related DOCX list-rendering bugs (plus a placeholder-path variant), reproduced against the real render pipeline:

1. **Numbering after `<br>` is lost.** `<br>` was only ever an *inline* soft break, but the block parser splits only on real newlines — so a `1. …` after a `<br>` rendered as literal text. This also leaked into the **placeholder** path (the router uses `contains_block_markdown` + a "has newline" check, neither of which saw `<br>`).
2. **Numbering continuation after a heading is lost.** A continuing item like `3.`, blank-separated (normal in procedural filings / *procesní podání*), failed the `ordered_list_is_genuine` gate — which deliberately rejects a blank-separated non-`1` number to keep dates (`23.\n\n24.`) out of lists. The numbering engine already honored the start number via `startOverride`; the blocker was purely *detection*.

## Fix

- **`expand_br_to_block_breaks()`** (`patterns.py`): promotes `<br>` to a real newline **only** when a resulting segment is a genuine block (heading / list item). Prose `<br>` stays a soft break; table rows and code fences are untouched; idempotent. Applied in `process_markdown_content` (base tool) and in the placeholder router (`dynamic_docx_tools.py`).
- **Running ordered counter** (`markdown_processor.py` + `block_elements.py`): tracks where the most recent top-level ordered list left off, preserved across headings + blank lines and reset by other content. A numbered line that matches the running count starts a list and resumes via the existing `startOverride`. `ordered_list_is_genuine` stays pure/local, so a lone blank-separated number with no predecessor still renders as prose (date disambiguation preserved — no existing behaviour changed).

## Behaviour

`Výrok soudu:<br>1. První<br>2. Druhý` → `Výrok soudu:` paragraph + numbered list **1, 2** (base **and** placeholder).

```
## I. Úvod
1. První
2. Druhý
## II. Argumentace
3. Třetí
4. Čtvrtý
```
→ a single visual sequence **1, 2, 3, 4** (continuation is a fresh numbering instance with `startOverride=3`). A single continuing item after a heading also works. A `1.` restart still restarts.

**Preserved (tested):** prose `Line A<br>Line B` stays one soft-break paragraph; `<br>` in table cells stays in-cell; standalone dates / blank-separated `5.`/`6.` with no preceding list stay prose.

**Known limitation (intentional, unchanged):** a placeholder value that is, *in isolation*, a blank-separated list starting at a non-`1` number is textually identical to a date pair and stays prose; the same value without blanks, or with a preceding `1./2.`, renders as a list.

## Tests

Adds `tests/test_docx_list_continuation_and_br.py` (17 tests). Full docx suite: **337 passing**, ruff clean.

### DIFF
--- a/docx_tools/block_elements.py
+++ b/docx_tools/block_elements.py
@@ -165,7 +165,7 @@ def add_table_to_doc(table_data, doc, col_alignments=None, borderless=False,
 # ---------------------------------------------------------------------------
 def process_list_items(lines, start_idx, doc, is_ordered=False, level=0,
                        return_elements=False, number_styles=None, bullet_styles=None,
-                       base_indent=None):
+                       base_indent=None, ordered_run=None):
     """Process markdown list items with proper Word numbering.
     When *return_elements* is True the created paragraph XML elements are
     removed from the document body and returned so the caller can re-insert
@@ -205,6 +205,7 @@ def process_list_items(lines, start_idx, doc, is_ordered=False, level=0,
     num_resolved = False
     current_num_id = None
     items_emitted = 0
+    last_item_number = None  # last top-level ordered number, to extend ordered_run
     while i < n:
         original_line = lines[i]
         stripped_left = original_line.lstrip()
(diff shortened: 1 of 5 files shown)


dvejsada:
@claude review

PR #78, opened 2026-06-28
fix(auth): surface Authorization header for API-key middleware

dvejsada:
## Problem

A correctly configured `API_KEY` was rejected when the client authenticated with the **`Authorization`** header (`Bearer <key>` or plain). Only `x-api-key` worked.

## Root cause

FastMCP's `get_http_headers()` strips a set of "problematic" headers **by default**, and that set includes **`authorization`** (and `proxy-authorization`). The middleware called it bare:

```python
headers = get_http_headers() or {}   # Authorization already removed here
```

so the `Authorization` header was gone before `_extract_key()` ran. Demonstrated directly against the installed FastMCP:

```
OLD (default)   get_http_headers() -> {}                                  # authorization stripped
NEW (include=)  get_http_headers(include={"authorization", ...}) -> {"authorization": "Bearer secret-123"}
```

## Fix

```python
headers = get_http_headers(include={"authorization", "proxy-authorization"}) or {}
```

The `include=` parameter removes those names from the exclusion set so the header reaches the middleware. `x-api-key` was never excluded and keeps working.

## Why existing tests missed it

Every `on_request` test mocked `get_http_headers` to return the header dict directly, masking the real stripping. This PR adds regression tests that:
- run the **real** `get_http_headers` (faking only the underlying request) and assert an `Authorization: Bearer` request authenticates; and
- assert `on_request` opts `authorization` into the `include` set.

Both fail on the old code and pass with the fix.

## Testing
- **638 passed** (full suite; the one deselected test is the pre-existing network-only pptx image test).
- `ruff check .` clean.

## Risk
- Tiny, self-contained change to `middleware.py`. No behaviour change for `x-api-key`; restores intended behaviour for `Authorization`. Auth is still disabled entirely when `API_KEY` is unset.

---
_Generated by [Claude Code](https://claude.ai/code/session_01X6T73vVGLZjUQ8dJ2bXkLF)_

### DIFF
--- a/middleware.py
+++ b/middleware.py
@@ -83,13 +83,10 @@ class ApiKeyAuthMiddleware(Middleware):
     # Middleware hook
     # ------------------------------------------------------------------
     async def on_request(self, context: MiddlewareContext, call_next):
-        # NOTE: get_http_headers() strips a set of "problematic" headers by
-        # default — and that set INCLUDES `authorization` (and
-        # `proxy-authorization`). Without opting them back in via `include=`,
-        # an `Authorization: Bearer <key>` request arrives here with the header
-        # already removed, so only `x-api-key` would ever authenticate. Pass
-        # include={...} so the Authorization header is surfaced to us.
-        headers = get_http_headers(include={"authorization", "proxy-authorization"}) or {}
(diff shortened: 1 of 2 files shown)


dvejsada:
@claude review

---

PR #79, opened 2026-06-29
fix(docx): preserve <center>/<div align> alignment under placeholder layout inheritance

dvejsada:
## Summary

Alignment set in a placeholder value via `<center>` or `<div align="...">` was being lost when rendered through a template — specifically whenever the template's placeholder paragraph carried its own paragraph properties (an explicit alignment, or even just paragraph spacing).

## Root cause

The recent "propagate placeholder layout (alignment/indent/spacing)" change (`d44bd8c` / `f9da538`) added `_propagate_format_to_block`, which makes plain prose inserted at a placeholder inherit the placeholder paragraph's layout (e.g. a justified letter body stays justified). It does this by removing each produced paragraph's whole `<w:pPr>` and stamping the placeholder's layout over it.

The guard that protects deliberately-styled paragraphs only checked for a `<w:pStyle>` (headings, lists, quotes). A `<center>` / `<div align="...">` block is plain-styled prose carrying only a `<w:jc>` (alignment) and **no** `<w:pStyle>`, so it slipped through the guard and had its alignment clobbered by the placeholder's own alignment.

Observed before the fix (driving the real placeholder-replacement path):

| Value | Placeholder paragraph | Rendered (before) | Expected |
|---|---|---|---|
| `<center>…</center>` | left-aligned | LEFT ❌ | CENTER |
| `<div align="right">…</div>` | spacing only | default ❌ | RIGHT |
| `<center>…</center>` | justified | JUSTIFY ❌ | CENTER |

When the placeholder paragraph had no `<w:pPr>` at all, alignment survived — which is why the existing tests (built on bare paragraphs) didn't catch it.

## Fix

In `_propagate_format_to_block`, capture an explicit alignment on the produced paragraph before swapping its `<w:pPr>`, and re-apply it afterward via the python-docx API (so `<w:jc>` lands in valid schema order). A deliberate markdown alignment now wins, while plain prose still inherits the placeholder's layout — both behaviors coexist (a centered block stays CENTER, and a following plain line still inherits the placeholder's JUSTIFY).

## Tests

- Added `TestAlignmentDirectivesInPlaceholders` covering the three cases above, including the coexistence case (aligned block keeps its alignment, following prose inherits the placeholder layout).
- Full docx suites pass: **208 passed** (205 existing + 3 new).

---
_Generated by [Claude Code](https://claude.ai/code/session_01EnTcVGF5wmxPPqdi63QP4s)_

### DIFF
--- a/docx_tools/dynamic_docx_tools.py
+++ b/docx_tools/dynamic_docx_tools.py
@@ -196,11 +196,9 @@ def _propagate_format_to_block(doc, inserted, src_ppr, fmt) -> None:
         if ppr is not None and ppr.find(qn('w:pStyle')) is not None:
             continue
         if layout_ppr is not None:
-            # An explicit alignment the markdown set on this otherwise-plain prose
-            # (e.g. via <center> or <div align="...">) is deliberate, so it must
-            # survive the layout inheritance instead of being clobbered by the
-            # placeholder's own alignment. Capture it before swapping the pPr and
-            # re-apply afterwards (via the API, so <w:jc> lands in schema order).
+            # Preserve a deliberate markdown alignment (e.g. <center>, <div align>)
+            # across the pPr swap so it wins over the placeholder's own alignment;
+            # re-apply via the API so <w:jc> lands in schema order.
             own_alignment = para.alignment
             if ppr is not None:
                 elem.remove(ppr)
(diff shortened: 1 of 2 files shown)


dvejsada:
@claude please review this PR.

---

PR #80, opened 2026-06-29
feat(transport): env-configurable stateless streamable-http (default off)

zahangirbu:
Adds STATELESS_HTTP env flag (default unset → stateful). When set truthy ("1"/"true"/"yes"/"on"), FastMCP runs streamable-http in stateless mode: each request is independent, no Mcp-Session-Id lookup is performed, and any pod can serve any follow-up call. This is required to scale office-docs horizontally on EKS (replicas >= 2)

Default remains False so every existing deployment is unaffected. Both launch paths honor the flag: mcp.run() (admin-off) forwards via transport_kwargs to run_http_async(stateless_http=...), and the admin-mounted ASGI path passes it directly to mcp.http_app().

Verified against fastmcp==3.4.2: mcp.http_app(stateless_http=False/True) both succeed; mcp.run_http_async source confirms stateless_http forwarding. Existing tests untouched.

### DIFF
--- a/.env.example
+++ b/.env.example
@@ -48,6 +48,21 @@ RUN_BLOCKING_BY_ASYNCIO_THREAD_ENABLED=
 # executor instead of fanning out and starving each other.
 RUN_BLOCKING_MAX_WORKERS=
 
+# --- Streamable-HTTP session mode ---
+# Default (empty / "false" / "0" / "no" / "off"): STATEFUL.
+#   Each MCP client gets an Mcp-Session-Id minted by the server and that
+#   session lives in this pod's memory. All follow-up requests for that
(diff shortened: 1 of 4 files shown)


dvejsada:
@claude review

dvejsada:
@zahangirbu Could you please only address point 1 from the review? I am happy to merge right after.

zahangirbu:
Hi @dvejsada,
Thanks for your review.

Point # 1, is already available through built-in log in transport.py of FastMCP. See the following logs when it is run with stateless. 


[06/29/26 15:00:13] INFO     Starting MCP server 'MCP Office Documents' with transport 'streamable-http'    transport.py:304
                             **(stateless)** on http://0.0.0.0:8958/mcp

PR #81, opened 2026-06-30
fix(docx): recognise headings/lists inside <center>/<div align> blocks

dvejsada:
## Summary

A heading (or list) placed **inside** a multi-line `<center>` or `<div align="...">` block — or written inline as `<center># Heading</center>` — was rendered with its markdown markers as **literal text** (e.g. a literal `# Centered Heading` instead of a centered Heading 1).

## Root cause

The multi-line alignment-block renderer (`process_alignment_block`) ran **inline formatting only** on each inner line; it never checked for headings, lists, or any block markdown. So `#` / `-` markers inside the block were emitted verbatim. This is a longstanding limitation dating to the original block-markdown feature (`68ea0bf`), not a regression from the recent alignment/newline work.

Observed before the fix (both the base markdown tool and template placeholders):

| Input | Before | After |
|---|---|---|
| `<center>` … `# Title` … `</center>` | literal `# Title` ❌ | Heading 1, centered ✅ |
| `<div align="right">` … `## H` … `</div>` | literal `## H` ❌ | Heading 2, right ✅ |
| `<center># Title</center>` (inline) | literal ❌ | Heading 1, centered ✅ |
| List items inside a block | literal `- item` ❌ | aligned list paragraphs ✅ |
| Heading before/after a block | already worked ✅ | still works ✅ |

## Fix

- Route the alignment block's inner lines through the **full block pipeline** (`process_markdown_block`) so headings, lists, tables, etc. are recognised, then stamp the block's alignment on every produced paragraph.
- The inline `<center>…</center>` form likewise promotes a wrapped heading to a real heading.
- Removed the now-obsolete `process_alignment_block` (and its `_ALIGN_CLOSE_RE` import in `block_elements.py`); the logic now lives in `markdown_processor._process_alignment_block`, where the block pipeline is available.

### Precedence

The block alignment is applied as **direct paragraph formatting** (`<w:jc>`), which correctly overrides any alignment defined by the heading's paragraph *style*. A heading wrapped in `<center>` centers regardless of what `Heading 1` defines; a **bare** heading (no wrapper) gets no direct `<w:jc>` and still follows its style. Headings outside an alignment block are untouched.

## Tests

- Base tool (`test_docx_base.py`): headings inside center/right blocks, inline-wrapped heading, and a list inside a center block.
- Template path (`test_docx_templates.py`): heading inside center and div-right placeholder content.
- Full docx suites pass: **214 passed** (208 prior + 6 new).

---
_Generated by [Claude Code](https://claude.ai/code/session_01EnTcVGF5wmxPPqdi63QP4s)_

### DIFF
--- a/docx_tools/block_elements.py
+++ b/docx_tools/block_elements.py
@@ -12,7 +12,6 @@ from .patterns import (
     UNORDERED_LIST_CAPTURE_PATTERN,
     _ALIGN_INLINE_RE,
     _ALIGN_OPEN_RE,
-    _ALIGN_CLOSE_RE,
 )
 from .inline_formatting import parse_inline_formatting
 from .patterns import _BR_RE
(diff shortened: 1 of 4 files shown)


dvejsada:
@claude please review this PR.

---

PR #82, opened 2026-07-14
Emit flat schemas for optional dynamic-tool args so descriptions survive

dvejsada:
## Problem

Optional args (`required: false`) of dynamic docx/email template tools were typed `Optional[...]`, which pydantic renders in the JSON schema as `anyOf: [{type}, {"type": "null"}]` with the description as a **sibling** of `anyOf`. Several MCP clients normalize `anyOf` schemas and silently drop that sibling description — so every optional argument reached the model **with no description at all**, while required args (flat `{type, description}` schemas) kept theirs.

Observed in production: a template with 26 args where exactly the `required: false` args arrived description-less; flipping an arg to `required: true` + default "fixed" it, which confirmed the pattern.

## Fix

Optionality is now expressed by the default alone: fields keep their **plain type** (pydantic does not validate defaults, so a `None` default with a plain type is fine) and are simply absent from the schema's `required` list. No `anyOf` is emitted anywhere.

Applies to:
- dynamic docx template args (`docx_tools/dynamic_docx_tools.py`),
- dynamic email template args and the email base fields `to`/`cc`/`bcc` (`email_tools/dynamic_email_tools.py`).

Behavioral note: explicitly passing `null` for an optional arg is now rejected instead of accepted — omit the arg instead (models do this naturally).

## Tests

New `tests/test_dynamic_args_schema.py` asserts for both docx and email registration that every arg (required, optional with default, optional without default, optional bool) has a flat `type` + `description` with no `anyOf`, and that the `required` list contains exactly the truly required args. Full suite: 649 passed (1 network-dependent pptx image test deselected).

https://claude.ai/code/session_01SYQSko3quHt3XWKx77E7HS

---
_Generated by [Claude Code](https://claude.ai/code/session_01SYQSko3quHt3XWKx77E7HS)_

### DIFF
--- a/docx_tools/dynamic_docx_tools.py
+++ b/docx_tools/dynamic_docx_tools.py
@@ -690,10 +690,16 @@ def _register_single_template(mcp: FastMCP, spec: Dict[str, Any],
         # Handle regular types
         py_type = TYPE_MAP.get(str(arg.get("type", "string")).lower(), str)
         required = bool(arg.get("required", True))
-        field_type = py_type if required else Optional[py_type]  # type: ignore[index]
         default = arg.get("default", (... if required else None))
         desc = arg.get("description", "")
-        fields[arg_name] = (field_type, Field(default, description=desc) if desc else default)
+        # Always use the plain type — never Optional[...]. Optionality is
+        # expressed by the default alone (the field is then absent from the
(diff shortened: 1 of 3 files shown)


dvejsada:
@claude review

Issue #67, opened 2026-06-19
[Bug] Numbered list restart


dvejsada:
Restart numbered list if 1 is used again on same list level.

PR #83, opened 2026-07-14
Fix nested-table placeholders and numbering continuation in docx renderer

dvejsada:
Two generic fixes for the dynamic DOCX template renderer.

## 1. Placeholders inside nested tables are now replaced

`_replace_placeholders_in_table` iterated `cell.paragraphs`, which does not descend into tables nested inside a cell (a common layout device in letterhead headers). Placeholders in nested tables leaked verbatim (`{{...}}`) into the generated document. The fix recurses into `cell.tables` with the same inline-only semantics as the outer table pass.

## 2. Ordered-list numbering continuation survives quotes and style-directive blocks

The running ordered-list counter (which lets numbered paragraphs resume across section headings, issue #67 follow-up) was reset by **any** non-heading block. Legal filings routinely interpose an evidence note (a `<!-- style: … -->`-styled paragraph) or a citation between numbered paragraphs — after such a block, a following `3.` failed the date-disambiguation check and rendered as plain prose instead of continuing the list.

Block quotes and comment/directive lines (plus the block a directive styles, which is consumed within the same dispatch) are deliberate interruptions of a numbered run, like headings — they no longer reset the counter. Prose paragraphs still reset it, so date disambiguation is unchanged for ordinary text.

## Tests

- `tests/test_docx_templates.py`: placeholder replacement inside a nested table.
- `tests/test_docx_list_continuation_and_br.py`: continuation across a blockquote and across a `<!-- style: … -->` styled block (asserted on numbering XML `startOverride`).

Full suite: 650 passed (1 network-dependent pptx image test deselected).

https://claude.ai/code/session_01SYQSko3quHt3XWKx77E7HS

---
_Generated by [Claude Code](https://claude.ai/code/session_01SYQSko3quHt3XWKx77E7HS)_

### DIFF
--- a/docx_tools/dynamic_docx_tools.py
+++ b/docx_tools/dynamic_docx_tools.py
@@ -489,6 +489,11 @@ def _replace_placeholders_in_table(
                 # Note: We don't pass doc to avoid inserting lists in table cells
                 _replace_placeholders_in_paragraph(paragraph, context, doc=None,
                                                    style_map=style_map)
+            # Tables nested inside a cell (layout tables in letterheads etc.)
+            # carry placeholders too; cell.paragraphs does not descend into them.
+            for nested in cell.tables:
+                _replace_placeholders_in_table(nested, context, doc=None,
+                                               style_map=style_map)
 
 
 def _replace_placeholders_in_document(doc: DocxDocument, context: Dict[str, str],
--- a/docx_tools/markdown_processor.py
+++ b/docx_tools/markdown_processor.py
@@ -138,16 +138,21 @@ def process_markdown_content(doc, content, return_elements=False,
             continue
         # --- All other block elements: delegate to block processor ---
         # Continuation survives headings, blank lines (above), block quotes and
-        # comment lines (directives attach to the block they style; both are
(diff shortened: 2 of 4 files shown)


dvejsada:
@claude review, namely focus on any unintended consequences of the parser changes

dvejsada:
Addressed the review in 5d376fb:

- **Malformed comment edge case — fixed.** The reset exemption now requires a complete `<!-- ... -->` line; an unclosed `<!--` renders as literal prose and resets the run like any prose paragraph (pinned by `test_unclosed_comment_line_is_prose_and_breaks_continuation`).
- **Directive breadth — confirmed intentional (option a).** The exemption deliberately keys off the directive, not the block type it styles: by styling a block the author explicitly marks it as special interposed content, and the residual date-misread risk (a following number that happens to exactly match the running count) is the same one the existing heading exemption already accepts. Inspecting the produced block type would couple the reset logic to the dispatcher's internals for a marginal edge case. Documented in the code comment and pinned by `test_directive_exemption_covers_any_styled_block`.

Full suite: 652 passed (1 network-dependent pptx test deselected).

---

Issue #66, opened 2026-06-19
[Enhancement] Custom style list


dvejsada:
For the custom templates, allow mapping of custom styles to replace built-in styles.

In addition, add style tag to allow use of custom style outside of regular mapping.

PR #84, opened 2026-07-15
Resolve list numbering and indents from the applied style

dvejsada:
## Summary

Three related fixes for ordered-list rendering with custom styles (follow-up to #66/#67), addressing reports that the `<!-- style: ... -->` directive and `style_mapping` overrides lose the style's numbering format and left/hanging indents.

1. **Numbering resolves from the style actually applied.** `resolve_ordered_abstract_num_id` is replaced by `resolve_ordered_numbering(doc, level, style_name)`, which tries the effective style first (directive or `style_mapping` override, following the `basedOn` chain and honoring the style's declared `numId`/`ilvl`), then the built-in `List Number/2/3` for that level, then the existing decimal-scan and synthesized fallbacks. The resolved `ilvl` is clamped to levels the abstract actually defines — previously nested items pointed at `ilvl 1/2` of single-level abstracts, which Word renders unnumbered.

2. **Style indents survive the direct `numPr`.** A direct `w:numPr` makes Word treat the numbering level's `w:ind` as direct formatting, silently overriding the paragraph style's own indents — including in the shipped default template (styled 576 twips, rendered 432). The renderer now re-asserts the applied style's `w:ind` as direct formatting on ordered items; attributes the style doesn't define (typically the number's `hanging`) still come from the numbering level, since `w:ind` attributes merge individually.

3. **The style directive styles lists through the style map.** When `<!-- style: X -->` targets a list, X is injected as the level-0 style in a per-block `StyleMap` instead of being stamped on every produced paragraph. Nested items keep `List Number 2/3` / `List Bullet 2/3` instead of being flattened, and a directive-styled numbered list restarts at 1 using the directive style's own numeral format and indents. A numbered line that renders as prose (e.g. a standalone date) still gets the style.

Also fixes a latent crash in the decimal-abstract fallback scan (`w:abstractNum` has no registered oxml class, so `w:`-prefixed xpath on it raises `XPathEvalError`), and documents the directive's exact scope and syntax rules in the Readme.

## Testing

- 13 new tests in `tests/test_docx_style_numbering.py`: custom-style abstract resolution (mapped and directive), `basedOn` inheritance, nested lists in single-level-abstract templates, missing-style fallback, indent re-assertion (and its absence when the style defines none), default-template indent regression, nested style preservation for numbered and bullet lists, non-genuine numbered lines, and the detached/placeholder path.
- Two existing resolver tests updated to the new API.
- Full suite: 668 passed; the only failure is a pre-existing PPTX test that downloads an image from the internet (unavailable in the sandbox, unrelated).
- Verified end-to-end on a simulated legal template: level-0 items carry the custom style, a restart instance of its upperRoman abstract, and the style's own indent restated directly; nested items keep `List Number 2` with a defined numbering level; a second directive block restarts at 1.

https://claude.ai/code/session_01T8KnbPTAZ5m6aSq2uZKkaP

---
_Generated by [Claude Code](https://claude.ai/code/session_01T8KnbPTAZ5m6aSq2uZKkaP)_

### DIFF
--- a/Readme.md
+++ b/Readme.md
@@ -294,7 +294,7 @@ Nesting and combinations work, e.g. `**bold with *italic* inside**`, `**~~bold s
 This paragraph uses the "Callout" style from your template.
 ```
 
-The `<!-- style: Name -->` directive applies a style to the next block only (every item of a list, or the table). Unknown styles fall back to the default with a warning. To remap styles globally or per template, see [Custom Templates](#-custom-templates).
+The `<!-- style: Name -->` directive applies a style to the next block only — one paragraph, one heading, the whole table, or the top-level items of a list (nested items keep the `List Bullet 2/3` / `List Number 2/3` styles). On a numbered list the style's **own numbering** is used: the list restarts at `1.` with the style's numeral format and indents. The directive must be alone on its line (no trailing spaces after `-->`), and the name must match the Word style name exactly. Unknown styles fall back to the default with a warning. To remap styles globally or per template, see [Custom Templates](#-custom-templates).
 
 </details>
 
--- a/Readme.md
+++ b/Readme.md
@@ -294,7 +294,7 @@ Nesting and combinations work, e.g. `**bold with *italic* inside**`, `**~~bold s
 This paragraph uses the "Callout" style from your template.
 ```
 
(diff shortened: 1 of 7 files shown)


dvejsada:
@claude review

dvejsada:
Addressed the review in 3bc4fc8:

- **🔴 `_find_decimal_abstract`** — correct catch, the fix hadn't been mirrored there. Now uses qualified-name `findall` like `_abstract_defined_ilvls`, with a regression test that deletes the built-in numbered styles to force the priority-3 fallback and verifies end-to-end rendering (the test reproduces the `XPathEvalError` against the old code).
- **Readme trailing-spaces note** — the dispatcher does strip, but a directive line ending in *two or more* spaces was caught earlier by the soft-line-break branch in `process_markdown_content` and rendered as literal prose. Fixed the trap itself: complete `<!-- -->` comment lines are now exempt from the soft-break branch (tests for both directive and non-directive comments), and the Readme caveat is dropped.
- **Style-map override** — now only the tuple matching the target list type (ordered vs unordered) is overridden.

Full suite: 671 passed; the one failure remains the pre-existing network-dependent PPTX image test.

---

dvejsada:
@claude review

PR #88, opened 2026-08-07
fix(xlsx): correctness, robustness and reference-semantics fixes for the Excel tool

dvejsada:
Reimplements the parts of #65 that are usage-agnostic bug fixes, plus a set of independent issues found while reviewing the Excel tool. **No new runtime dependencies** — `formulas`/numpy/scipy, the Alpine→Debian base-image switch and the recalculation feature are all deliberately excluded (see *Not included* below).

## Breaking change

`B[n]` in a formula now resolves against **the row the formula is on**, not the current table's first data row.

```
=B[0]*C[0]  @row 4 → =B4*C4    (was =B4*C4)
            @row 5 → =B5*C5    (was =B4*C4)   ← every row was identical
            @row 6 → =B6*C6    (was =B4*C4)
```

Previously `B[n]` and `T1.B[n]` resolved to the same cell, so the short form was a redundant alias — and wrong in the way that mattered: every data row of a computed column produced the same formula, silently breaking running totals, per-row products and growth rates. The tool description already documented `B[0]` as a current-row reference; this makes the behaviour match. `T1.B[n]` remains the way to name one fixed cell.

There is no stored markdown to migrate. Anyone writing `B[1]`/`B[2]` as *table* indices should switch to `T1.B[1]`/`T1.B[2]`.

## Correctness fixes

| Fix | Before | After |
|---|---|---|
| `number` type, European decimal | `1,5` → **`15.0`** | `1.5` |
| `types:` comma inside a format | `number:#,##0.00` split into 2 columns, shifting all later ones | parsed intact |
| Formula in a typed column | written verbatim → `#NAME?` | resolved, given the column's format |
| Accounting negative | `($1,234)` → text | `-1234.0` |
| Cross-sheet `SUM` range | `=SUM(S!B4:S!D4)` | `=SUM(S!B4:D4)` (canonical) |
| Percent precision | `50.5%` displayed as **`51%`** | `0.0%` → `50.5%` |
| Grouped decimals | `1500.75` displayed as **`1,501`** | `#,##0.00` → `1,500.75` |
| Negative thousands | unformatted (`>= 1000` test) | grouped |
| Missing table / unknown sheet | silent wrong reference | `WARNING` naming what exists |

## Files Excel would refuse to open

- **Duplicate or empty Table headers.** Excel requires every `tableColumn` name to be non-empty and unique (case-insensitively); openpyxl copies them from the header cells without validating, so a table with two `Q1` columns and `auto_filter=True` emitted duplicate entries and opened to a repair prompt. Blanks now get a positional name, duplicates a suffix — only when `auto_filter` builds a Table.
- **Formulas over Excel's 8192-character limit** make Excel reject the *whole workbook*, not the cell. Now stored as inline strings: one visibly wrong cell in a file that opens.

## Silently wrong output

- **Sheet names that aren't plain identifiers.** `John's Data` produced `'John's Data'!B2` — the quoted section ends at the apostrophe and the rest is garbage. Apostrophes are now doubled, and the cross-sheet patterns accept Excel's quoted form on input, which is the only way a name containing an operator character can be unambiguous (`=A1-P&L!T1.B[0]` cannot be parsed; `=A1-'P&L'!T1.B[0]` can). Bare names are unaffected.
- **`nan` / `inf` / `1_000`.** `float()` accepts these; openpyxl writes non-finite values as an empty `<v>`, so the cell arrived **blank**. PEP 515 underscores turned a product code into a number. Both are text now.

## Circular-reference detection

New `xlsx_tools/circular_refs.py`. A formula that depends on itself makes Excel show a warning and resolve to 0 — invisible server-side. Dependency-graph analysis over the formula strings, iterative DFS (a long running-total column is a deep chain; recursion would hit Python's frame limit on ordinary workbooks). Diagnostic only: logs, never blocks. openpyxl only, no new dependency.

## Structural

`markdown_to_excel` and `_markdown_to_excel_buffer` were near-identical ~90-line duplicates that had already drifted — only one logged, only one warned about colliding sheet names — so the LibreChat artifact path ran a lesser version of the tool. Collapsed into one `_build_workbook()`, with the save and all diagnostics in the buffer function so both callers get them.

## Test suite

The suite was **49 failed / 334 passed / 29 collection errors** before this branch. It is now **879 passed, 0 failed, 0 errors**.

Three causes, none in the repo's code:
1. The local venv had a corrupted `fastmcp-slim` (509 of 534 files missing, including `fastmcp/__init__.py`), so `fastmcp` resolved to an empty namespace package. Environment-only — reinstalling fixed 42 failures and 15 errors. **Nothing to review here**, but worth knowing if others hit it.
2. `TestToolSignatureDefaults` read `main.py` without an encoding, so on Windows it used cp1252 and raised `UnicodeDecodeError` on the currency symbols in the tool descriptions. Passes on UTF-8 platforms, which is why CI never caught it. Fixed.
3. `test_create_eml_buffer` asserted raw bytes contain `b"Test Subject"`, but the header is RFC 2047 encoded (`=?utf-8?q?Test_Subject?=`) so the literal never appears. The email tool is correct; the test now decodes the header.

163 new tests across three files.

## Not included from #65

- **Formula recalculation.** Pulls `formulas`, numpy, scipy and 4 more packages, forces the Docker base image from Alpine to Debian slim, and `formulas`/`schedula` are **EUPL-1.1** (copyleft) against this repo's MIT. Deferred pending that decision.
- **Financial-modeling mode** — already removed from #65 at your request.
- **`default_font`** — belongs with the broader cell-styling discussion.

Two notes for @andormarkus if #65 continues:
- `Dockerfile` still runs `apk upgrade` on both stages after switching to `python:3.12-slim`. `apk` doesn't exist on Debian; the image cannot build. Reported in review round 2, still unfixed.
- The structured-reference filtering in `formula_engine.py` doesn't work. Its comment says `Table1[Revenue]` yields `BLE1` which `_is_valid_coord` rejects — but `BLE1` is column 1669, row 1, and perfectly valid, so the phantom edge survives. This branch strips structured and external-workbook references before matching instead.


dvejsada:
@claude review

dvejsada:
Thanks — the header-collision finding is real and I reproduced it before fixing. Fixed in a09cf9b.

**`_ensure_unique_table_headers` duplicate (blocking)** — confirmed exactly as described. The blank branch was an `if`/`elif`, so taking the blank path skipped the uniqueness check entirely:

```
["Column2", ""]      -> ["Column2", "Column2"]              *** duplicate ***
["Column3", "", ""]  -> ["Column3", "Column2", "Column3"]   *** duplicate ***
```

Rather than duplicating the suffix loop into the blank branch, both cases now run through one path — build a candidate name (the heading, or the positional fallback if blank), then suffix until unique. The positional fallback isn't special; it's just another candidate:

```
["Column2", ""]      -> ["Column2", "Column2_2"]            ok
["Column3", "", ""]  -> ["Column3", "Column2", "Column3_2"] ok
["Q1", "Q1", "Q1"]   -> ["Q1", "Q1_2", "Q1_3"]              unchanged
["", ""]             -> ["Column1", "Column2"]              ok
```

Six regression tests added, including `["Q1", "Q1", "Q1_2"]` → `["Q1", "Q1_2", "Q1_2_2"]`, where a real heading already looks like a generated suffix — the case that breaks a naive "append _2" fix. The warning now also fires only when a heading actually changed, rather than whenever a blank was seen.

**Text-stored formulas in `circular_refs`** — taking this one rather than leaving it. You're right that the docstring tolerates spurious edges, but this isn't a parsing approximation: the cell is one we *deliberately* made non-executable, so reporting it as a cycle is just wrong. Verified `data_type` distinguishes the two reliably after a reload (`'f'` vs `'s'`, both with a `"="`-prefixed value), and keyed the check on that. Two tests: a text-stored `=A1+1` is not a cycle, a real one still is.

**Column-width formula detection** — agreed, folded into `_strip_markdown_formatting` so there's one definition of "is this a formula" instead of an ad-hoc `lstrip('*\`')` alongside it.

On the unverified test claim: the suite runs green locally — **885 passed, 0 failed, 0 errors** (up from 879 before this commit; it was 49 failed / 334 passed / 29 errors on master). Three of those causes are covered in the PR description; the largest was a corrupted `fastmcp-slim` in the local venv rather than anything in the repo, which is worth knowing if you hit it too.

dvejsada:
@claude review

PR #89, opened 2026-08-07
feat(xlsx): cell styling primitives and template-defined named styles

dvejsada:
The usage-agnostic alternative to #65's financial-modeling mode, per your review comment there: *"give the MCP server general tools to achieve the desired outcome — e.g. ability to set background or font color to a cell, or set custom excel template with styles."*

Two mechanisms, both general-purpose. The server takes no view on what a style should mean.

## `<!-- styles: ... -->` directive

```markdown
<!-- styles: A[-1]:D[-1]=bg:darkblue;color:white;bold, D[3]=bg:yellow;bold -->
| Month | Sales | Costs | Profit        |
|-------|-------|-------|---------------|
| Jan   | 100   | 60    | =B[0]-C[0]    |
| Feb   | 200   | 90    | =B[0]-C[0]    |
| Mar   | 300   | 120   | =B[0]-C[0]    |
| Total | =SUM(T1.B[0]:T1.B[2]) | =SUM(T1.C[0]:T1.C[2]) | =SUM(T1.D[0]:T1.D[2]) |
```

Comma-separated `<target>=<attributes>` entries; attributes separated by `;` (`,` and `:` are both already taken).

| Target | Meaning |
|---|---|
| `B2` | Absolute worksheet cell |
| `B[0]` | Table-relative — first data row; `B[-1]` is the header |
| `B2:D5`, `C[0]:C[3]` | Rectangular block, either form |

| Attribute | Effect |
|---|---|
| `bg:<colour>` | Background fill |
| `color:<colour>` | Font colour |
| `bold`, `italic`, `underline` | Weight and decoration |
| `style:<Name>` | Named style from your template |

Colours are 6-digit hex (`FFFF00`, `#FFFF00`) or one of 18 names (`gray`/`grey` both work).

**On the bracket notation.** In a formula `B[0]` means "the row I am on" — meaningless in a directive, which sits on no row. A directive belongs to exactly one table, so here it anchors to that table instead. Same notation, an unambiguous anchor in each context.

## Named styles from your own template

Drop an `.xlsx` with named cell styles at `custom_templates/custom_xlsx_template.xlsx` and every style in it becomes referenceable as `style:<Name>`. Resolved through `template_utils`, so it follows the same custom→default search path as the docx, pptx and email templates. A house look lives in one file instead of hex codes repeated across every document.

Styles are registered into the workbook up front, which is what lets an unknown `style:Name` be reported at write time rather than shipping a workbook that quietly lost its formatting. openpyxl's built-in `Normal` is excluded, since re-registering it conflicts.

## Behaviour worth knowing

- **Styling wins.** It runs as a final pass over the finished table, so an explicit instruction beats the built-in header fill, the formula fill and inline `**bold**`.
- **It only sets colours and font flags.** Values, number formats, borders and alignment are untouched — verified by a test that a `currency:$` column keeps `$#,##0.00` and its parsed value through styling.
- **Unmentioned attributes are preserved.** `bg:yellow` on an already-bold header does not un-bold it.
- **Ranges expand to the full rectangle**, not the corners.
- **Failures never cost you the document.** A mistyped colour, an unresolvable target or an oversized range is logged and skipped; an oversized range is *dropped* rather than silently truncated to something that looks like it worked.

## Tests

66 new tests in `tests/test_xlsx_styling.py`. Full suite **951 passing**, ruff clean. The Readme example above is executed in CI-equivalent form and asserted to produce exactly what it documents.

Not included: restyling the built-in header/formula colours from the template. That needs a naming convention (`Header`, `Formula`) and is worth deciding separately — the current scope keeps the template purely additive.


dvejsada:
@claude review

dvejsada:
Good catch — this is real, and it was worse than the diff suggested. Fixed in 63c6e0c.

Reproduced against openpyxl directly. `cell.style = name` replaces font, fill, border, alignment **and** number format together, so a template style declaring only a font wiped three properties the renderer had already set:

```
before                    number_format='$#,##0.00'  border.left='thin'  align='right'
after style:Plain         number_format='General'    border.left=None    align=None
```

**I didn't take the suggested fix, though** — capturing and unconditionally restoring all three would be wrong in the other direction. A `NamedStyle` legitimately carries a number format; that's part of what a cell style *is* in Excel. Restoring blindly would mean a template style deliberately declaring `0.00%` could never apply it:

```
after style:Rich (declares 0.00%)   number_format='0.00%'   ← must survive
```

So each property is restored only where the style left it at its **default** — precisely the case where the style expresses no preference. `number_format == 'General'`, `border == Border()`, `alignment == Alignment()`. No need to hunt down the `NamedStyle` object either: reading the values off the cell *after* application gives the style's own values.

```
style:Plain (font only)      -> number_format='$#,##0.00'  border='thin'  align='right'  bold=True
style:Rich  (declares 0.00%) -> number_format='0.00%'      border='thin'  align='right'  bold=True
```

Five tests: font-only style preserves number format; preserves border and alignment; a style declaring `0.00%` still wins over a `currency:$` column; a style declaring `center` still applies; and `style:Total;bg:yellow` combined with `types: currency:$` keeps all three behaviours at once.

**You were also right that the documentation overstated the guarantee.** "Never changes a cell's number format" was true of `bg:`/`color:`/`bold` but not of `style:`. The Readme and tool description now separate the two: colour and font attributes touch nothing else, while a `style:` reference applies whatever the template's style declares. That distinction is the honest contract and it's now the tested one.

Full suite **956 passing**, ruff clean. On the note about not being able to execute the suite: that's worth flagging on the review side generally — a static read caught this one because the clobbering is visible in the API contract, but the failure only *manifests* at runtime. Anything that needs execution to surface won't show up in that pass.

dvejsada:
@claude review

dvejsada:
@claude review

dvejsada:
Both passes came back clean, and the only open item is the `number_format='General'` edge case raised in each. I dug into it — **it isn't a trade-off that was skipped, it's not solvable**, and I want to correct that on the record before it stands as a to-do.

Both reviews suggest it could be fixed by "introspecting the `NamedStyle` definition itself rather than diffing before/after cell state." That doesn't work. In the object, "unset" and "explicitly `General`" are the same value:

```
Plain  (font only)              number_format='General'   builtinId=None
Reset  (declares 'General')     number_format='General'   builtinId=None
```

And it isn't lost in openpyxl's model either — it never reaches the file. Both styles serialise to byte-identical entries, with no `applyNumberFormat` attribute to separate them:

```xml
<cellStyleXfs>
  <xf numFmtId="0" fontId="1" fillId="0" borderId="1"/>   <!-- Plain, xfId=1 -->
  <xf numFmtId="0" fontId="1" fillId="0" borderId="1"/>   <!-- Reset, xfId=2 -->
  <xf numFmtId="10" fontId="1" fillId="0" borderId="1"/>  <!-- Pct,   xfId=3 -->
</cellStyleXfs>
```

So no reader recovers the intent, because the intent was never written. The current restore-at-default approach isn't a lesser approximation of a better available method — it's equivalent to the best one that exists. Clearing an inherited format would need an explicit directive attribute (a future `format:General`), not a smarter lookup.

Recorded in the `_apply_named_style` docstring in 64138ba so this doesn't get re-attempted as a "proper fix". No behaviour change; **956 passing**, ruff clean.

One process note, since it's now three passes: each one states it couldn't execute the suite or linter. That's fine for what static reading catches — the clobbering bug in the first pass was genuinely visible in the API contract, and finding it was worth the round. But the repeated "everything else checked out" is a read, not a verification, and re-running it against an unchanged commit (the last two passes were both on `63c6e0c`) mostly re-derives the previous conclusion. If another pass is wanted, pointing it at something specific would get more out of it than a general re-review.

PR #127, opened 2026-09-18
pptx: draw a title the layout cannot hold instead of dropping it (#118)

### LINKED ISSUE #118: pptx: blank slides drop title without a way to keep it
`role_for_slide()` maps `blank` to `ROLE_BLANK`, and `classify_layout()` only
returns `ROLE_BLANK` for a layout with no meaningful placeholders
(`pptx_tools/layouts.py`, `if not types: return ROLE_BLANK`). A blank layout
therefore can never have a title placeholder, so any `blank` slide carrying a
`title` loses it and gets:

```
slide N: layout 'Prázdný' has no title placeholder; the title '...' was dropped.
```

The schema advertises `title` on every slide type, including `BlankSlide`, so
this is a field that is accepted, validated, and then discarded — reliably, not
occasionally.

**Suggested fix:** on a `blank` slide, draw `title` as a text box at the
position the template's own title placeholder uses (readable from the `content`
role layout), rather than dropping it. Failing that, reject `title` on
`BlankSlide` at validation time so the failure is loud instead of a warning
nobody reads.

**Impact:** every author who reaches for `blank` hits this once. It is the only
guaranteed-on-every-run warning we found.

dvejsada:
Closes #118.

## The problem

A Blank layout has no title placeholder — that is what makes it blank — and `role_for_slide()` sends every `blank` slide (and an untitled `quote`) there. So a `blank` slide carrying a `title` had it accepted by the schema, validated, and then discarded with a warning on every single run:

```
slide N: layout 'Prázdný' has no title placeholder; the title '...' was dropped.
```

## The fix

`_apply_title()` now falls back to drawing the title itself when the resolved layout has no placeholder to fill, rather than dropping it:

- **Where.** `LayoutResolver.title_style()` reads the first title placeholder in the template, preferring the body layouts (`content`, `title_only`, two-column, `image_text`, `section`) over the cover layout, whose title is a centred block halfway down the slide. The answer is cached, so a deck of blank slides walks the layouts once.
- **How it looks.** `placeholder_style.read_title_style()` takes that placeholder's geometry (python-pptx resolves it through the master, so a layout with no `<a:xfrm>` still answers) plus the `<a:defRPr>` of the first paragraph properties in the chain *layout placeholder → master placeholder → the master's `<p:titleStyle>`*. That element is copied onto the run wholesale rather than re-derived attribute by attribute, so theme references (`+mj-lt`, `schemeClr`) survive and keep following the theme. Alignment and vertical anchoring come from the same chain.
- **Fit.** `apply_autofit()` replaces python-pptx's default `<a:spAutoFit>` on a new text box, so a long title shrinks instead of resizing the band.

The title is drawn before a blank slide's own elements, so anything the author positions still draws on top.

Only a template whose layouts carry no title placeholder at all still warns — the box then goes in a band across the top of the slide, which is a guess worth saying out loud.

## Files

| File | Change |
|------|--------|
| `pptx_tools/placeholder_style.py` | New. `read_title_style()`, `draw_title_box()`, `title_placeholder()` |
| `pptx_tools/layouts.py` | `TITLE_REFERENCE_ROLES`, cached `LayoutResolver.title_style()` |
| `pptx_tools/slide_builder.py` | `_apply_title()` draws instead of dropping; `_default_title_style()` |
| `main.py`, `docs/powerpoint-slides.md`, `docs/development/tools/powerpoint.md` | Tool description, the blank-slide paragraph, and the pipeline diagram, module map and a new "Titles on a layout that has none" section |
| `tests/` | The three tests that asserted the drop now assert the draw; `tests/test_pptx_title_fallback.py` covers geometry, inherited style, autofit, ordering and the untitled-template warning |

## Testing

`pytest -m "not network"` — 1706 passed. `ruff check .` — clean.

---


dvejsada:
@claude review

---

dvejsada:
Thanks — both observations checked, one of them fixed in 8db3d5c.

**Multi-line title:** fixed rather than left as-is. `draw_title_box()` now applies the alignment and the copied `<a:defRPr>` to every paragraph and every run, not just `paragraphs[0]`, so a title containing a newline does not drop its second line to python-pptx's default face at half the size. The copy-and-retag moved into a small `_run_properties()` helper now that it runs per run, and `test_every_line_of_a_multi_line_title_is_styled` covers it. Worth doing here even though `_set_title()` has the same shape: on the placeholder path later paragraphs inherit from the layout's list style, so only the text-box path actually loses the styling.

**Degenerate `0` width/height:** agreed, leaving it. A template with a zero-sized title placeholder is already visibly broken in PowerPoint itself, and a guard would only replace one wrong-looking result with another.

`pytest -m "not network"` — 1707 passed; `ruff check .` clean.

---

PR #128, opened 2026-09-18
pptx: draw inside the template's content area, not the whole slide (#119)

### LINKED ISSUE #119: pptx: Blank-role slides assume an empty canvas and draw over layout decoration
Related to #1 but distinct. Anything routed to `ROLE_BLANK` is positioned from
the top-left of the slide, on the assumption that a blank layout is visually
empty. Corporate templates routinely keep logo, rules and footer on their blank
layout — ours does.

Reproduction: an untitled `quote` (which `role_for_slide()` deliberately sends
to `ROLE_BLANK`) renders its text across the template's gold and red header
rules:

```json
{ "type": "quote", "text": "Citace bez nadpisu.", "attribution": "§ 1" }
```

The quote text lands on the gold rule at ~21% height and the attribution on the
red rule at ~29%. Same for user-positioned `blank` elements — authors have to
discover by trial that nothing may sit above roughly `y: "22%"`.

**Suggested fix:** derive a usable content rectangle for blank-role slides
instead of using the full slide. The bounding box of the `content` role
layout's body placeholder is a good approximation and needs no new template
metadata. Alternatively expose the safe area so callers can position against it.

dvejsada:
Closes #119.

## The problem

A layout with no body placeholder — Blank, Title Only — sent everything the builder draws into a hardcoded band starting 1.5 inches from the top-left of the slide, on the assumption that a blank layout is an empty canvas. It is not: templates keep a logo, a header rule and a footer band on it.

Two visible symptoms:

- an untitled `quote` (which `role_for_slide()` deliberately sends to the blank role) rendered across the template's header rules;
- `kpi` and `timeline` slides use the `title_only` role, which has no body placeholder either, so they drew from 1.5in — **inside** the shipped template's title band, which runs to 24.6% of the slide height (1.85in).

## The fix

The template already says where content goes; read it instead of guessing.

- `placeholder_style.read_content_rect()` reads a layout's **largest** body placeholder — the same "largest, not first" rule `_add_title_content_slide()` already applies to a slide's own placeholders, so a Comparison layout's heading strip is not mistaken for its content box.
- `LayoutResolver.content_area()` caches the first rectangle it finds, preferring the single-content layouts (`content`, `section`) over a two-column layout, whose body is half the slide.
- `SlideHelpers._content_area()` uses it. The hardcoded band survives only as the last resort, for a template with no body placeholder on any layout.

On the shipped 16:9 template an untitled quote moves from `top: 20%` to `top: 26.6%`, the content layout's own body top.

## Blank slides keep absolute coordinates

The issue offers "derive a content rectangle" or "expose the safe area"; for the author-positioned `elements` of a `blank` slide it has to be the second. Re-basing those coordinates would silently relayout every existing deck and contradict the documented meaning of `x: 0`.

So the band is published instead: `validate_templates()` reports `content_area` as percentages, and `list_presentation_templates` passes it through with `include_layouts`, in the exact form a caller writes into an element:

```json
{"content_area": {"layout": "Nadpis a obsah",
                  "x": "6.9%", "y": "26.6%", "w": "86.2%", "h": "63.4%"}}
```

Both tool descriptions now point at it.

## Files

| File | Change |
|------|--------|
| `pptx_tools/placeholder_style.py` | `Rect`, `content_placeholders()`, `read_content_rect()` |
| `pptx_tools/layouts.py` | `CONTENT_REFERENCE_ROLES`, cached `LayoutResolver.content_area()` |
| `pptx_tools/helpers.py` | `_content_area()` replaces the inline fallback rectangle |
| `pptx_tools/templates.py` | `_content_area_summary()`; `content_area` in the report |
| `main.py` | `content_area` in the listing; blank-slide and listing tool descriptions |
| `docs/` | Content-placement section, module map and test table on the dev page; the blank-slide paragraph in `powerpoint-slides.md`; the listing section of `templates.md` |
| `tests/test_pptx_content_area.py` | New: placement for quote/kpi/timeline, the reading rules, the last-resort band, and the reported percentages |

## Testing

`pytest -m "not network"` — 1717 passed. `ruff check .` — clean.

---


dvejsada:
@claude review

---

dvejsada:
Both nits taken, in `b202bd1` — they're in code this PR already touches, so they belong here rather than in a follow-up.

- **The unreachable guard.** Confirmed: `PowerpointPresentation` is the only class `SlideHelpers` is mixed into, and it assigns `self._layouts` in `__init__` before `_build_slides()` runs, so the `getattr(self, "_layouts", None)` branch could never be taken. `_content_area()` now calls `self._layouts.content_area()` directly, and the mixin declares `_layouts` alongside `presentation` in its class-level hints and docstring, which is how it already states its dependency on the consuming class.
- **The duplicated BODY/OBJECT pair.** `_content_placeholders()` now delegates to `placeholder_style.content_placeholders()`, so the tuple exists once. No import cycle: `placeholder_style` imports nothing from the package.

`pytest -m "not network"` — 1717 passed; `ruff check .` clean.

---

PR #129, opened 2026-09-18
pptx: fill the template's picture placeholder on image slides (#120)

### LINKED ISSUE #120: pptx: ROLE_IMAGE_TEXT is unreachable — PICTURE placeholders are never filled
`ROLE_IMAGE_TEXT` is defined, classified and given a positional fallback in
`pptx_tools/layouts.py`, but:

- no entry in `SLIDE_TYPE_ROLE` requests it — `"image"` maps to `ROLE_CONTENT`;
- `PP_PLACEHOLDER.PICTURE` appears nowhere outside `layouts.py`, so no builder
  ever populates a picture placeholder.

`_build_image_slide()` calls `_add_image()` into a computed rectangle on the
content layout instead. The result is that a template's photo layouts are dead
weight: ours has two (`Foto partnera`, `Obrázek s titulkem`) that the tool will
never select and could not fill if it did.

**Suggested fix:** have `image` prefer `ROLE_IMAGE_TEXT` when the template
offers it, and insert into the `PICTURE` placeholder so the template's own
crop, frame and position apply. Fall back to the current drawn-rectangle path
when no such layout exists.

dvejsada:
Closes #120.

## The problem

`ROLE_IMAGE_TEXT` was defined, classified and given a positional fallback, but nothing reached it: no entry in `SLIDE_TYPE_ROLE` asked for it, and `PP_PLACEHOLDER.PICTURE` appeared nowhere outside `layouts.py`. A template's photo layouts were dead weight — the shipped template has one (`Obrázek s titulkem`) the tool would never select and could not have filled.

## The fix

- **Reach the role.** `role_for_slide(slide, resolver)` returns `ROLE_IMAGE_TEXT` for an `image` slide when `resolver.provides(ROLE_IMAGE_TEXT)`. `provides()` counts a *configured* name as well as a detected one, so a template that maps `image_text:` to a layout the signature rules deliberately refuse to classify still gets it. This is the one role that depends on the template rather than on the slide alone, which is why it is resolved in `role_for_slide()` next to the existing `two_column`/`comparison` and untitled-quote decisions.
- **Fill the placeholder.** `_build_image_slide()` puts the image in the slide's `PICTURE` placeholder via `_fill_picture_placeholder()`, which uses python-pptx's `insert_picture()` — so the template's own frame, crop and position apply instead of a rectangle the builder invented. `body` and `caption` go into the layout's body placeholder (on a picture layout that placeholder *is* the caption area), the caption as a trailing italic paragraph.

On the shipped 16:9 template an image slide now looks like this, all three shapes being the layout's own placeholders:

```
Title 1               placeholder   'Chart'
Picture Placeholder 2 placeholder   <image>
Text Placeholder 3    placeholder   'Up and right\nFig 1'
```

## What still uses the computed rectangle

The old path is the fallback, not a leftover:

- the template has no picture layout;
- or the layout resolved to has a picture placeholder but **no body placeholder** to hold this slide's `body`/`caption` — text is never dropped in order to keep the nicer frame.

An unfilled picture placeholder is removed in both cases, as is an unused body placeholder, so no slide ships with "Click icon to add picture" waiting in it.

## Files

| File | Change |
|------|--------|
| `pptx_tools/layouts.py` | `role_for_slide(slide, resolver)`; `LayoutResolver.provides()` |
| `pptx_tools/helpers.py` | `_fill_picture_placeholder()`, `_remove_placeholder()` |
| `pptx_tools/slide_builder.py` | `_build_image_slide()` picture-layout path; `_fill_picture_layout()`, `_add_caption_paragraph()` |
| `main.py`, `docs/powerpoint-slides.md`, `docs/development/tools/powerpoint.md` | The IMAGES line, the images paragraph, and a new "Pictures in a picture placeholder" section plus the layout-role paragraph and test table |
| `tests/test_pptx_picture_layout.py` | New: role choice, the filled placeholder, text placement, the failed-image path, both fallbacks |
| `tests/test_pptx_slide_types.py`, `tests/test_pptx_schema.py` | The side-by-side tests now name the content layout — the path they were always exercising; the inline-image test asserts on the image rather than the shape type, which differs between the two paths |

## Testing

`pytest -m "not network"` — 1728 passed. `ruff check .` — clean.

---


dvejsada:
@claude review

---

dvejsada:
Both taken, in `52720cb`.

**The missing overflow warning** was the right catch — and you're right that it lands hardest here, since a picture layout's text area is the smallest one a template offers. `_fill_picture_layout()` now goes through `_fit_text()` like every other placeholder-filled body, so it both shrinks and warns past the shrink floor. The reason it didn't before was the caption: `_fit_text()` measures `bullets` alone. Rather than add a second code path, the caption is measured as one more item (`bullets + [Bullet(text=caption)]`) — it is set at caption size, smaller than body text, so the estimate errs generous rather than under-warning. `test_overfull_text_is_shrunk_and_reported` covers it.

**The stale docstring** on `_content_slide()` no longer lists `image`, with a line saying why: `_build_image_slide()` calls `_new_slide()` and `_apply_title()` itself so it can look for a `PICTURE` placeholder in between.

`pytest -m "not network"` — 1729 passed; `ruff check .` clean.

---

PR #130, opened 2026-09-18
pptx: report each layout's detected role from the listing tool (#121)

### LINKED ISSUE #121: pptx: list_presentation_templates should report role coverage
`include_layouts: true` returns layout names, but not which of the eight roles
each one satisfies, nor which roles nothing satisfies. A template that is
missing a role fails silently at build time with a positional fallback.

Concrete case: in both our templates `Záhlaví oddílu` carried only a title
placeholder, so it classified as `title_only` rather than `section`. Two
consequences, neither visible without reading the source:

1. every `section` slide warned and fell back to layout index 2;
2. because `_by_role` keeps the first match, `title_only` resolved to
   `Záhlaví oddílu` (index 2) ahead of `Jenom nadpis` (index 5) — so **`kpi` and
   `timeline` slides were built on the section-divider layout**, putting a
   centred title through the middle of the KPI figures.

Diagnosing this took an hour of reading `layouts.py` and running
`classify_layout()` by hand. The classifier already exists; surfacing its output
would have made it a single call.

**Suggested fix:** add to each layout entry the role it classifies as (or
`null`), plus a top-level `unmatched_roles` list and the resolved
role → layout mapping. Consider a standalone `validate_presentation_template`
tool for the admin UI, since customers upload their own templates.

dvejsada:
Closes #121.

## The problem

`list_presentation_templates` returned layout *names*, so a template that resolves oddly could not be diagnosed from the outside. The classifier's answer existed all along — `LayoutResolver.describe()` was written for this tool and never wired to it, the same shape of gap as #120.

## The fix

`validate_templates()` now reports `resolver.describe()` as `layouts`, and the tool passes it through:

```json
{"index": 2, "name": "Záhlaví oddílu", "role": "title_only", "vertical": false,
 "placeholders": ["TITLE", "DATE", "FOOTER", "SLIDE_NUMBER"]}
```

Placeholder types are included because they say *why* a layout classified as it did — the classifier reads types and nothing else.

Two smaller changes in the same area:

- **`missing_roles` is always present**, empty list included. A model cannot tell an absent key from "nothing is missing", and this is the field that says a slide type will land on a fallback layout. `unknown_configured_layouts` is passed through too; the report already carried it.
- **Roles are tracked per layout, not per name.** `describe()` read `self._roles_of[layout.name]`, so two layouts sharing a name reported the same role for both — the last one classified. They now live in a list parallel to `_layouts`.

## The reported case, in one call

The issue's template — a section layout carrying only a title — now reads:

| Field | Value |
|-------|-------|
| `layouts[2]` | `{"name": "Záhlaví oddílu", "role": "title_only", …}` |
| `roles.section` | `null` |
| `roles.title_only` | `"Záhlaví oddílu"` |
| `missing_roles` | `["section", …]` |

Which is the whole diagnosis, including the part nothing warned about: that layout is also what `kpi` and `timeline` slides get, because it comes before the real title-only layout. `test_a_section_layout_detected_as_title_only_is_visible` pins exactly that.

## On the suggested `validate_presentation_template` tool

Not added. `admin/analysis.py` already analyses an uploaded template in more detail than a tool response should carry — placeholder signatures per layout, theme fonts and colours, embedded slide count, warnings — and the admin UI previews a real deck through the template before saving. A second MCP tool would duplicate it with less. Happy to add one if you want the same diagnostics available to a model outside the admin UI.

## Files

| File | Change |
|------|--------|
| `pptx_tools/layouts.py` | `describe()` reports per-layout roles from a list parallel to the layouts; compact placeholder types |
| `pptx_tools/templates.py` | `report["layouts"]` is `describe()`, not names |
| `main.py` | Pass-through, `missing_roles` always present, tool description and docstring |
| `docs/templates.md`, `docs/development/tools/powerpoint.md` | What each diagnostic field answers, and how the reporting is built |
| `tests/test_pptx_templates.py` | Per-layout roles, the mis-detected section layout, duplicate names, and the tool boundary |

## Testing

`pytest -m "not network"` — 1734 passed. `ruff check .` — clean.

---


dvejsada:
@claude review

---

dvejsada:
Both fixed in `552ede4`.

**The duplicated docs section** was exactly what you diagnosed — the table went in above the existing paragraph instead of into it. The two are now one section: `content_area` is described once in the table row, and the paragraph below it keeps only the part the row doesn't cover (where it is read from, and why the tool's own slides stay inside it).

**`missing_roles` on a template that would not open** — I took this one rather than leaving it, because this PR is what made the key unconditional, so the misleading reading is new. A template that failed to open was never classified, so an empty list beside its `error` is a verdict nobody reached; the tool now returns `error` alone in that case, and `test_a_template_that_will_not_open_reports_only_the_error` pins it.

On the `idx` drop: agreed, and thanks for checking the repo for readers of the old shape — the classifier reads placeholder *types* only, so types are what explain a detected role. `idx` still matters for the two-column and footer invariants, but those are builder-side, not something the listing answers.

`pytest -m "not network"` — 1735 passed; `ruff check .` clean.

---

PR #131, opened 2026-09-18
pptx: make the warnings channel data rather than prose (#122)

### LINKED ISSUE #122: pptx: Warnings are prose strings, not structured data
`self.warnings` collects English sentences. A caller can display them to a human
and nothing more — it cannot branch on them, count them by kind, or act.

**Suggested fix:** return objects with a stable `code`, `slide` index,
`severity`, and a human-readable `message`; keep the current strings as the
`message` field for compatibility.

This unlocks the obvious follow-on: the advice in nearly every overflow warning
is "split this across more slides", which the tool could offer to do itself
behind an opt-in flag.

dvejsada:
Closes #122.

## The problem

`self.warnings` collected English sentences. A caller could show them to a human and nothing else — not branch on them, not count them by kind, and above all not tell a deck that *lost* content from one that was merely shrunk.

## The shape

Each entry is now a `SlideWarning` (new `pptx_tools/warnings.py`):

```json
{"code": "image_failed", "severity": "error", "slide": 0,
 "message": "image could not be loaded (404 Not Found); a placeholder was drawn."}
```

| Field | Meaning |
|-------|---------|
| `code` | stable identifier — the thing to branch on |
| `severity` | see below |
| `slide` | index in the caller's `slides` list, counting from zero, exactly as the message has always said; absent for a deck-wide warning |
| `message` | the sentence, unchanged |

**Severity is about the deck, not about how unusual the situation is:**

- `error` — something the caller asked for is not in the file (dropped subtitle, image that would not load, element skipped for being off-slide);
- `warning` — it is there, but shrunk, clamped, or laid out on a layout that was not the obvious one;
- `info` — a substitution the caller probably does not mind (a format or template picked for them, a table font reduced a point or two).

So `any(w["severity"] == "error" for w in warnings)` answers "did this deck lose anything", which is the question the prose could not be asked.

Severity is derived from the code through `WARNING_SEVERITY`, so **one code always means one severity** — a caller that learns `image_failed` is an error must never meet one that is not. `test_every_code_has_a_severity` fails if a code is added without an entry. All 29 warning sites were given codes.

## Compatibility

`str(warning)` renders the line the channel used to hold, and `PowerpointPresentation.warning_messages` returns those lines. That is what the server log, the admin preview header (`X-Preview-Warnings`, which goes to a person) and the existing string assertions use — so the prose did not have to change anywhere to gain the structure.

The tool response is the one deliberate shape change: `warnings` is now a list of objects rather than strings, which is what the issue asks for. A clean deck still returns a bare URL string, unchanged.

## Not included

The auto-split follow-on the issue mentions ("the tool could offer to do it itself behind an opt-in flag"). That changes what the tool *produces*, not what it reports, and needs its own decision about when splitting is welcome — `text_overflow` and `table_overflow` are now the codes it would key off.

## Files

| File | Change |
|------|--------|
| `pptx_tools/warnings.py` | New. `SlideWarning`, the codes, `WARNING_SEVERITY`, `make_warning()` |
| `pptx_tools/slide_builder.py` | `_warn(index, code, message)`, `_warn_deck(code, message)`, `warning_messages`; codes at all 29 sites |
| `pptx_tools/base_pptx_tool.py`, `main.py` | Types, and `as_dict()` on the way out |
| `admin/preview.py`, `admin/app.py` | The preview hands back rendered lines — its header is read by a person |
| `docs/` | The warnings paragraph in `powerpoint-slides.md` with the object shape and a field table; the channel section, module map and test table on the dev page; `testing.md`; `adding-a-tool.md` |
| `tests/test_pptx_warnings.py` | New: records, the taxonomy, codes in practice, and the tool boundary |
| other pptx tests | String assertions moved to `warning_messages`, otherwise unchanged |

## Testing

`pytest -m "not network"` — 1752 passed. `ruff check .` — clean.

---


dvejsada:
@claude review

---

dvejsada:
Good question on `table_overflow` — I checked the path rather than reasoning from the message, and `warning` is correct.

`_build_table_slide()` calls `_create_styled_table()` unconditionally, which creates **every** row; `table_overflows()` then only asks whether the rendered table needs more vertical space than the content rectangle has, at the font size `fit_table_font_size()` already shrank it to. Nothing is dropped — all the rows are in the file, the table just extends past the box. That is "there, but not as asked", which is the `warning` definition.

The contrast that makes the line concrete is `timeline_detail_dropped`, which *is* an `error`: `_timeline_detail_band()` returns `(0, 0)` and the detail text frames are never created, so that content genuinely is not in the file.

The message's "split it across slides" is advice about legibility, not a report of loss — which is exactly the kind of thing prose was bad at conveying and the `severity` field now settles. Leaving it as is; no code change.

---

PR #132, opened 2026-09-18
pptx: give bullets in a text box the master's glyphs (#123)

### LINKED ISSUE #123: pptx: Bullets placed beside an image render without bullet glyphs
In `_build_image_slide()`, when `body` is present the text goes into
`slide.shapes.add_textbox(...)` rather than a placeholder, so it inherits no
list formatting from the master. `_fill_bullets()` writes the indent levels but
the glyphs never appear — the same markdown that shows bullets on a `content`
slide shows plain lines here.

**Suggested fix:** apply the master's list style explicitly on the generated
text box, or reuse a content placeholder where the layout provides one.

dvejsada:
Closes #123.

## The problem

Bullets beside a picture or a chart go into `slide.shapes.add_textbox(...)`, not a placeholder. A text box inherits its paragraph formatting from the presentation's `<p:defaultTextStyle>`, which has no bullets — so `_fill_bullets()` wrote the levels and the glyphs never appeared:

```xml
<a:p><a:pPr algn="l"/><a:r><a:t>one</a:t></a:r></a:p>
<a:p><a:pPr algn="l" lvl="1"/><a:r><a:t>two</a:t></a:r></a:p>
```

The identical markdown on a `content` slide bulleted correctly, because a body placeholder inherits from the master's `<p:bodyStyle>`.

## The fix

`placeholder_style.apply_list_style()` applies that same body style explicitly, per paragraph level:

```xml
<a:pPr algn="l" marL="228600" indent="-228600">
  <a:buFont typeface="Arial" …/><a:buChar char="•"/>
</a:pPr>
```

Three decisions worth stating:

- **Only the list formatting is copied** — `marL`, `indent` and the bullet definition. Size, colour and line spacing stay as the builder set them: the body style's 28pt first level is meant for a full-width placeholder, not a column beside a picture.
- **Each paragraph takes its own level's style**, so a nested bullet is indented as it would be in a placeholder rather than all levels sharing level 1's indent.
- **`<a:pPr>` is rebuilt in schema order** — bullet properties after `lnSpc`/`spcBef`, before `defRPr`. PowerPoint reports a file whose paragraph properties are out of order as damaged, and a paragraph that carries a font size already has a `defRPr` to sit before. `test_it_keeps_the_children_of_pPr_in_schema_order` pins it.

A master with no `<p:bodyStyle>` returns False and the text box is left as it was; a deck on such a template still builds.

## One helper instead of two copies

The image and chart builders had the same four lines each (add box, fill bullets, autofit). Both now call `_add_bulleted_textbox()`, so the glyph step cannot be forgotten by the next builder that puts text beside something.

## Files

| File | Change |
|------|--------|
| `pptx_tools/placeholder_style.py` | `body_list_levels()`, `apply_list_style()`, `_replace_bullet_properties()` |
| `pptx_tools/slide_builder.py` | `_add_bulleted_textbox()`, used by the image and chart builders |
| `docs/development/tools/powerpoint.md` | Content-placement section, module map, test table |
| `tests/test_pptx_bullet_glyphs.py` | New: glyphs beside a picture and a chart, per-level indents, sizing left alone, placeholders still inheriting, `<a:pPr>` order, replacing an existing `buNone`, and the no-body-style template |

Nothing under `docs/` described the old rendering — the user reference says only that `chart` and `image` take a `body` "to put takeaways beside them", which is what now happens.

## Testing

`pytest -m "not network"` — 1761 passed. `ruff check .` — clean.

---


dvejsada:
@claude review

---

dvejsada:
On the dropped-child observation: stays as is, and the schema is why. `CT_TextParagraphProperties` allows exactly `lnSpc`, `spcBef`, `spcAft`, the four bullet groups, `tabLst`, `defRPr`, `extLst` — `_BEFORE_BULLETS` and `_AFTER_BULLETS` cover every one of those except the bullet groups, which are the ones being replaced on purpose. A child outside that set is not valid OOXML in the first place, so there is nothing real to preserve; if one ever showed up, keeping it would carry an invalid element into the output rather than save anything.

Thanks for checking the ordering against the schema sequence rather than taking the comment's word for it — that was the part most likely to be quietly wrong.

---

PR #133, opened 2026-09-18
pptx: column widths, cell fills and merged cells on table slides (#124)

### LINKED ISSUE #124: pptx: Table formatting options are minimal
`TableSlide` exposes `align`, `header_color`, `zebra` and `font_size` only.
Missing, in rough order of how often we wanted them:

- column widths (currently distributed evenly, so a long description column and
  a two-character column get equal space);
- per-cell or per-row fill, for flagging severity in a risk table;
- merged cells, for grouped headers.

dvejsada:
Closes #124.

All three of the issue's asks, in its order of usefulness.

```json
{"type": "table", "title": "Risk register",
 "rows": [["Risk", "Impact", "Owner"],
          ["Data loss", "High", "Ops"],
          ["Latency", "Low", "Eng"]],
 "widths": [3, 1, 1],
 "fills": [{"row": 1, "col": 1, "color": "C00000"}, {"row": 2, "color": "accent2"}],
 "merges": [{"row": 0, "col": 1, "col_span": 2}]}
```

## `widths` — relative, not absolute

`[3, 1, 1]` means "three times the space", which is what a caller can state without knowing what content rectangle the layout will hand the table. The weights are normalised to the table's own width and the last column takes the rounding remainder, so the columns still add up exactly — `test_the_columns_still_add_up_to_the_table` compares against an unweighted table.

## `fills` — one cell or a whole row

`{row, col, color}`, with `col` omitted meaning the whole row. Applied after the header and zebra passes so an explicit fill wins, and colours take hex or a theme name like every other colour in the schema.

## `merges` — before the text, not after

`Cell.merge()` **concatenates the text of the cells it joins**, so merging after filling gives a block reading `"Impact\nOwner"`. Merges are therefore applied to the empty table first, and cells that come back `is_spanned` are skipped when the text is written. A 1×1 merge is a schema error rather than a silent no-op, since it reads like an intention.

## Everything is checked against the real table first

None of these can fail a deck:

| Situation | Result |
|-----------|--------|
| `widths` length ≠ column count | dropped whole, reported — guessing which column a short list meant is worse than leaving them equal |
| fill or merge outside the table | that one skipped, reported |
| merges overlapping each other | later one skipped, reported; the first to claim a cell keeps it |

Overlap detection lives in the builder rather than being left to python-pptx, which raises part-way through a block — that would take the whole deck with it. Three new warning codes (`table_widths_ignored`, `table_fill_ignored`, `table_merge_ignored`), all `warning` severity: the table is drawn, just not styled as asked.

## Files

| File | Change |
|------|--------|
| `pptx_tools/schema.py` | `CellFill`, `CellMerge`, and `widths`/`fills`/`merges` on `TableSlide` |
| `pptx_tools/helpers.py` | `_create_styled_table()` takes the three extras; `_set_column_widths()` |
| `pptx_tools/slide_builder.py` | `_table_widths()`, `_table_fills()`, `_table_merges()` — the validation and the warnings |
| `pptx_tools/warnings.py` | The three codes and their severities |
| `main.py`, `docs/powerpoint-slides.md`, `docs/development/tools/powerpoint.md` | A TABLES line in the tool description, the user-facing table paragraph and type row, and a "Table formatting" section with the merge-order gotcha |
| `tests/test_pptx_table_formatting.py` | New: proportions and rounding, per-cell/row/theme fills, fills beating zebra, spans and untouched rows, and every rejection path |

The published flat schema stays flat with the two new nested models — the generated JSON has no `oneOf`, `$ref` or `discriminator`, which I checked directly.

## Testing

`pytest -m "not network"` — 1779 passed. `ruff check .` — clean.

---


dvejsada:
@claude review

---

dvejsada:
Both observations checked; no changes.

**The unreachable `default` in `resolve_fill(fill.color, TABLE_ALT_ROW_FILL)`** — correct, and it stays. `resolve_fill()` reaches its default only when `color is None` or when `parse_color()` fails, and neither can happen here: `CellFill.color` is required and the `Color` type's validator has already accepted only 6-digit hex or a theme name, so `parse_color()` cannot fall through either. Passing `None` instead would contradict the signature, which types `default` as `RGBColor` rather than `Optional[RGBColor]` — the call would then be the one place that violates it. A real constant of the right type is the honest argument for an argument that is never consulted.

**`Optional[int]` on `CellFill.col`** — thanks for checking rather than pattern-matching the rule. It is the dynamic-template tools that must not use `Optional[...]`, because their arguments are published to clients whose schema dialects mishandle it; `CellFill` is a static model inside the slide union, where `col` genuinely means "omit this to fill the whole row".

---

PR #134, opened 2026-09-18
pptx: measure text against real font metrics, not character counts (#125)

### LINKED ISSUE #125: pptx: Autofit estimates from character counts rather than font metrics
The fill estimate is derived from text length, then written into the deck as an
autofit instruction that PowerPoint recomputes on open. `Pillow` is already a
dependency and could measure real glyph widths for the deck's actual fonts,
which would make the shrink warnings trustworthy rather than indicative.

dvejsada:
Closes #125.

## The problem, in numbers

The estimate divided the character count by one mean glyph width, the same for every face and every letter. Two bullets of **identical length** in a 6×3in box at 18pt:

| Text | Old (character count) | Now (measured) |
|------|----------------------|----------------|
| `"iii " × 80` | 0.71 | **0.41** |
| `"WWW " × 80` | 0.71 | **1.22** |

One comfortably fits, the other overflows, and the old rule could not tell them apart. `test_character_counting_could_not_tell_them_apart` pins that it used to.

## Which face gets measured

`pptx_tools/text_metrics.py` resolves a font file through Pillow — already installed, since python-pptx depends on it — best case first:

1. **the deck's own typeface**, read from the theme's minor latin font (`theme_body_typeface()`);
2. **a metric-compatible substitute** — Carlito for Calibri, Liberation Sans/Arimo for Arial, Liberation Serif/Tinos for Times New Roman, Caladea for Cambria, Gelasio for Georgia. Their advances are identical by design, so measuring one *is* measuring the real thing;
3. **any installed sans face** — a real measurement of the wrong font, still shape-aware and still far closer than a character count;
4. **nothing** — `measure_lines()` returns None and the old `AVG_CHAR_WIDTH_RATIO` arithmetic runs unchanged, so a bare environment behaves exactly as before.

The runtime image installs one package per family in that table — `font-carlito`, `font-liberation`, `font-caladea`, `font-gelasio` — which makes case 2 the usual one, and `test_every_metric_compatible_face_is_installed` reads the Dockerfile so the table and the image cannot drift apart again. Case 1 is out of reach for the shipped templates: they set **Aptos**, which has no free metric-compatible clone — worth knowing, since the issue's phrasing ("the deck's actual fonts") is only achievable where the actual font is installable. Every Alpine package name was checked against the package index rather than assumed.

Long words are now broken across lines instead of counted as one, which was the other way the old estimate under-measured.

## Files

| File | Change |
|------|--------|
| `pptx_tools/text_metrics.py` | New. Font resolution, substitutes, wrapped line counting, theme face |
| `pptx_tools/helpers.py` | `estimate_text_fill()` measures, falls back to arithmetic |
| `pptx_tools/slide_builder.py` | Reads the deck's face once, passes it to both fit paths |
| `pptx_tools/constants.py` | `BULLET_INDENT_INCHES`; the section preamble and the ratio now describe the fallback role |
| `Dockerfile`, `requirements.txt` | The four font packages; Pillow named explicitly, since this imports it directly |
| `docs/` | Fit-estimation section rewritten, the "approximate" known limitation restated as what *is* still approximate, module map, test table, and a clause in the user-facing warnings paragraph |
| `tests/test_pptx_text_metrics.py` | New: glyph sensitivity, wrapping, long-word breaking, face preference order, caching, the no-font fallback, the deck's own face, and the Dockerfile/table agreement |

## Testing

`pytest -m "not network"` — 1800 passed. `ruff check .` — clean. The Docker font install is the one thing I could not exercise here; the package names are verified to exist in Alpine, but the build itself runs in CI.


dvejsada:
@claude review

---

dvejsada:
Both fixed in `c737c8c`.

**The unreachable substitutes** — you caught a real gap, and it was wider than Georgia: `Caladea` for Cambria was equally unavailable, since `font-carlito` and `font-liberation` carry neither. I installed them rather than dropping the entries: Alpine has `font-caladea` and `font-gelasio` (checked against the package index, as with the first two), and in a repository whose documents are set in Cambria and Georgia those are the two substitutes most worth having exact.

The underlying problem is that a table in Python and a package list in a Dockerfile can drift silently, which is exactly what happened. `test_every_metric_compatible_face_is_installed` now reads the Dockerfile's `apk add` line and fails if `METRIC_COMPATIBLE` ever promises a family the image does not carry, with a companion test that every family has a known package mapping at all. The Dockerfile comment says so too, so the next person adding a substitute sees why the list is there.

**The stale preamble** — right, and it contradicted the comment two lines below it. It now describes what those constants actually are: the fallback ratio for an environment with no font file, plus the line height and shrink floor that the measured path uses as well.

`pytest -m "not network"` — 1800 passed; `ruff check .` clean.

---

PR #135, opened 2026-09-18
pptx: make inline markdown work in every text field, as the tool claims

dvejsada:
Found by a review pass over the pptx tool instructions: checking the claims the tool makes against what it does.

## The mismatch

The tool description says **"INLINE MARKDOWN in any text field"** and **"LINKS: [label](url) works in any text field"**. `docs/powerpoint-slides.md` said "every text field". Thirteen of them printed the markers instead:

| Worked | Printed `**bold**` literally |
|--------|------------------------------|
| body bullets, table cells, quote text, agenda items, closing subtitle and contact, blank-slide text and shape text | **title (every slide type)**, title-slide **subtitle**, quote **attribution**, KPI **value**, **label**, **delta**, timeline **label**, **detail**, image **caption**, two-column **heading**, **chart title**, **x/y axis titles** |

So a model that followed the instructions and wrote `*Case* **Alfa** v. Beta` as a title got asterisks on the slide, and `[docket](url)` rendered as its markdown source.

## The fix

Every one of those assigned `paragraph.text` directly. They now go through one entry point, `inline_formatting.write_text()`, which renders the grammar when the text carries any and assigns plainly when it does not. It also absorbs the four-line `needs_inline_processing` dance that the quote body, the image caption, `_fill_bullets()` and the table-cell writer were each doing by hand — so the rule genuinely lives in one place, which is what makes the new `AGENTS.md` line ("never `paragraph.text`") enforceable rather than aspirational.

**`draw_title_box()` had to change with it.** It stamped the template's title style onto each *run*, which would now overwrite a caller's bold and drop a link's `<a:hlinkClick>` — the blank-slide title was the one field that came out unformatted *and* unmarked in my first pass, which is how I caught it. The style is now the paragraph's `<a:defRPr>`: the same inheritance a real placeholder gives its runs, with run-level choices sitting on top. A side effect worth naming: a multi-line title is now one paragraph with `<a:br>` rather than two paragraphs, which matches the soft-break model the Word tool settled on in #110.

**Chart text needed one accommodation.** A hyperlink needs a relationship in the part that owns the run, and python-pptx cannot resolve a chart title's part — a link there raises `'ChartTitle' object has no attribute 'part'`. PowerPoint does not follow a link inside chart text anyway, so `write_text(..., hyperlinks=False)` renders the label and drops the target. The flag threads through `apply_inline_formatting()` and the recursive parser rather than being caught after the fact, so the one place that cannot have links declares it.

## Verified on the running server

Not just the suite — the deck a live `create_powerpoint_presentation` call produced:

```
slide 0  [('Case', I), (' ', ), ('Alfa', B), (' v. Beta', )]     ← title
         [('docket', →link)]                                      ← subtitle
slide 1  [('ARR', B)] [('+12%', I)]                               ← KPI label, delta
slide 2  [('Filing', B)] [('docket', →link)]                      ← timeline label, detail
slide 3  [('— ', B), ('Brown v. Board', B I)]                     ← attribution
slide 4  [('For', B)] [('Against', I)]                            ← column headings
```

## Two instruction corrections from the same pass

- The inline-markdown line and the user docs now **name the fields**, including chart titles, rather than saying "any text field" and being wrong about thirteen of them — and they state the chart-link exception.
- **`language`** claimed every text run is stamped. Chart text — its title, axis titles and category labels — is not reached by `_apply_language()`, so a Czech deck's chart is proofed in the viewer's language. Left as is by your decision; the parameter description and the docstrings now say so instead of promising more.

`AGENTS.md` gains both rules (write text with `write_text()`, never `paragraph.text`; every warning takes a code with a severity), since this is precisely the kind of thing the next change gets wrong by writing the obvious thing.

## Testing

`tests/test_pptx_inline_everywhere.py` parametrises **every** text field the schema has — reaching into chart parts as well as shapes and tables — three assertions each: no literal markers, the emphasis applied, the link a real hyperlink (or, in a chart, its label rendered). So a new field that forgets `write_text()` is a failing test. Plus the grammar's restraint (`5 * 3 * 2` stays arithmetic, `\*` loses its backslash) and the title-style interaction.

`pytest -m "not network"` — 1876 passed. `ruff check .` — clean.


dvejsada:
@claude review

---

dvejsada:
Good catch, and the claim in my description was the wrong one — fixed in `6f38858`, along with two more you didn't mention.

**The two you found** — `_fill_bullets()` and the table-cell writer — are converted. You're right that they were correct already and right that they contradicted the rule this PR adds; a rule with two exceptions in the tree it ships with is not a rule.

**Two more of the same kind**, found by taking your point seriously and grepping rather than trusting the list I'd already made: `chart_utils` wrote `chart_title`, `x_title` and `y_title` with `paragraph.text`. Those are caller strings, so a chart title printed its markers exactly like the ten fields this PR set out to fix — the claim would have been false again the moment it merged.

**Chart text needed one accommodation.** A hyperlink needs a relationship in the part that owns the run, and python-pptx cannot resolve a chart title's part: a link there raises `'ChartTitle' object has no attribute 'part'`. Since PowerPoint does not follow a link inside chart text anyway, `write_text()` now takes `hyperlinks=False`, threaded through `apply_inline_formatting()` and the recursive parser rather than caught after the fact — the one place that cannot have links declares it instead of discovering it. The tool description and the user docs state the exception; the dev page explains why it exists.

**On the weak assertion:** correct, the `or` made it unfailable — setting alignment alone creates the `<a:pPr>` the second branch looked for. It now reads the paragraph's own size and weight (`Pt(40)`, bold), which is what the docstring meant.

The test matrix reaches into chart parts now, so those three fields are covered by the same three assertions as the rest.

`pytest -m "not network"` — 1876 passed; `ruff check .` clean.

---

PR #137, opened 2026-09-19
word: render a continued numbered list as one Word list (#136)

### LINKED ISSUE #136: [Bug] A continued numbered list is a separate Word list, so it never renumbers
## What happens

When a numbered list resumes after interposed content (a heading, a paragraph, an evidence note, a table), the continuation gets the right visible numbers but is a **separate numbering instance**:

```markdown
1. First
2. Second

Some prose.

3. Third
4. Fourth
```

renders as

```xml
<w:num w:numId="17"><w:abstractNumId w:val="8"/>
  <w:lvlOverride w:ilvl="0"><w:startOverride w:val="1"/></w:lvlOverride></w:num>
<w:num w:numId="18"><w:abstractNumId w:val="8"/>
  <w:lvlOverride w:ilvl="0"><w:startOverride w:val="3"/></w:lvlOverride></w:num>
```

Items 1–2 carry `numId 17`, items 3–4 carry `numId 18` with a hard-coded `startOverride` of 3.

## Why it is wrong

The numbers are frozen at generation time. Adding an item to the first part in Word does not renumber the second part — the two runs drift out of sync, which for a numbered filing means wrong cross-references.

## Cause

Continuation is handled at the detection layer only: `_continues_ordered_run()` in `docx_tools/markdown_processor.py` decides that a lone `3.` is a list item, but rendering goes through a fresh `process_list_items()` call whose numbering state is local, so the first item always mints a new `<w:num>` with `startOverride=<the literal number>`. The running-count cell carries the number but not the identity of the numbering instance that produced it.

## Expected

The continuation should reuse the previous run's `numId`/`ilvl` (no second `startOverride`), which is what Word's own "Continue numbering" produces: one logical list that recomputes its numbers on edit. A list starting at `1.` must still restart on a fresh instance.

dvejsada:
Fixes #136.

## The bug

A numbered list resumed after interposed content got the right visible numbers but the wrong structure. `1. / 2.` → prose → `3. / 4.` produced two numbering instances:

```xml
<w:num w:numId="17"><w:abstractNumId w:val="8"/>
  <w:lvlOverride w:ilvl="0"><w:startOverride w:val="1"/></w:lvlOverride></w:num>
<w:num w:numId="18"><w:abstractNumId w:val="8"/>
  <w:lvlOverride w:ilvl="0"><w:startOverride w:val="3"/></w:lvlOverride></w:num>
```

The `3` is frozen at generation time, so adding an item to the first part in Word never renumbered the second — for a numbered filing, that means wrong cross-references after the first edit.

## Cause

Continuation was implemented at the detection layer only. `_continues_ordered_run()` knew a lone `3.` was a list item, but rendering went through a fresh `process_list_items()` call whose numbering state is local, so the first item always minted a new `<w:num>` with `startOverride=<the literal number>`. The running cell carried the number but not the identity of the instance that produced it.

## Fix

The running cell now carries the instance (`num_id`, `ilvl`, `style`) alongside `next`. A top-level list whose first item equals the running count re-attaches that `numId`/`ilvl` instead of minting a restart instance. Word counts every paragraph on the same `numId` as one list however much sits between them, so the parts renumber together — the same thing Word's own "Continue numbering" produces. The example above now yields one instance, items 1–4 all on `numId 17`.

Guards kept:

- a list opening at `1.` still restarts on a fresh instance;
- a continuation whose level-0 style differs gets its own instance, or it would silently inherit the earlier list's numeral format and indents;
- after a mid-list `1, 2, 1` restart, a later continuation resumes the *newest* instance;
- a run opening at `5.` keeps `startOverride=5`, and its continuation adds nothing.

## Scope

Reuse is tracked at level 0 only. A nested list resumed after interposed content still starts a fresh instance, so its numbers stay frozen — recorded under "Known limitations" in `docs/development/tools/word.md` rather than left silent.

## Docs

The "Ordered lists" section of `docs/development/tools/word.md` gains the continuation rule and the limitation; `numbering.py`'s module docstring states it only ever *starts* lists; `docs/markdown-reference.md` and the tool description in `main.py` now say the parts stay one list in Word.

## Tests

The eight tests that pinned the old `startOverride` design assert the shared instance instead, via a new `_one_list()` helper. Five cases are new: the regression itself (one instance, no second `startOverride`), a run opening above 1, restart-then-continuation, a differently styled continuation, and nested children not disturbing the parent run.

`ruff check .` clean; `pytest -m "not network"` 1881 passed.

https://claude.ai/code/session_01G2gn9rNFmDtaWTBsMovrLH

---
_Generated by [Claude Code](https://claude.ai/code/session_01G2gn9rNFmDtaWTBsMovrLH)_


dvejsada:
@claude review


---

dvejsada:
Thanks — CI is green on this head ([Lint & Test #79](https://github.com/ForLegalAI/mcp-ms-office-documents/actions/runs/35424147629): `ruff` clean, 1881 passed), so the static trace matches what actually runs.

On the nit: the `item_number != 1` clause stays, and it is not unreachable — `cont_number` can be `1`, because `ORDERED_LIST_PATTERN` accepts `0.` and the run records `last_item_number + 1`. A list ending on a `0.` item sets the count to 1:

```markdown
1. jedna
0. nula

Proza.

1. znovu
```

With the clause, the trailing `1.` restarts on a fresh instance (`numId 18`, `startOverride 1`); with it removed, that same input reuses `numId 17` and the list never restarts, contradicting the documented rule that a list beginning again at `1.` restarts. Both halves verified by running the renderer. So the line is load-bearing on that path rather than decorative — leaving it exactly as is.


---

PR #138, opened 2026-09-19
docker: install only font packages Alpine actually has

dvejsada:
Unblocks the next release. **v4.0-beta.4 published no Docker image** — [that run failed](https://github.com/ForLegalAI/mcp-ms-office-documents/actions/runs/35387644063), so `georgx22/mcp-office-docs:v4.0-beta.4` does not exist and the pull command in its release notes does not work.

## The failure

```
#17 [runtime 3/9] RUN apk add --no-cache font-carlito font-liberation font-caladea font-gelasio
#17 0.507 ERROR: unable to select packages:
#17 0.507   font-caladea (no such package)
#17 0.507   font-gelasio (no such package)
```

Alpine packages neither Caladea nor Gelasio, in `main` or `community`, on any current branch — checked against the APKINDEX for `latest-stable`, `v3.22` and `v3.21`. `apk add` is all-or-nothing, so the two missing names took `font-carlito` and `font-liberation` down with them and failed the build. The line arrived with the font-metrics work (#134) and has never built successfully; beta.4 was the first release to run it.

## The fix

The image installs `font-carlito` and `font-liberation`, which exist and whose files are the stems `load_font()` looks for (verified by unpacking the two `.apk`s: `Carlito-Regular.ttf`, `LiberationSans-Regular.ttf`, `LiberationSerif-Regular.ttf`, `LiberationMono-Regular.ttf`).

Caladea and Gelasio stay in `METRIC_COMPATIBLE`. The table is right about the faces — it just is not a statement about Alpine, and a Debian host with `fonts-crosextra-caladea` / `fonts-gelasio` measures Cambria and Georgia exactly. What changes is that in the container those two now reach `GENERIC_SANS` — a real measurement of the wrong font, the tier `text_metrics` already documents — instead of being promised a file that was never in the image.

## Why CI stayed green through this

`test_every_metric_compatible_face_is_installed` compared the Dockerfile against a hand-written face → package map, so a package that exists nowhere satisfied it. Both directions are covered now:

- the map records Caladea and Gelasio as unpackaged (`None`), and the families that costs are pinned by name, so losing another is a deliberate edit rather than a passing test;
- a package the table never names is flagged as dead weight in the image;
- **`test_every_font_package_exists_in_alpine`** (`network`-marked) resolves every font package the Dockerfile names against Alpine's real APKINDEX. It fails on the old line and passes on this one — the check that was missing, in the one place CI can see what `apk` will see.

## Docs

The fit-estimate section of `docs/development/tools/powerpoint.md` now says which families the container cannot measure and why; `docs/deployment.md` tells a host operator that installing the two Debian packages *beats* the image rather than matching it.

## Verification

`ruff check .` clean; `pytest -m "not network"` 1883 passed; `pytest -m network` 6 passed. No Docker daemon in this environment, so the build itself is not run here — the package names are verified against the same index `apk` resolves against, and the font files by unpacking the packages.

https://claude.ai/code/session_01G2gn9rNFmDtaWTBsMovrLH

---
_Generated by [Claude Code](https://claude.ai/code/session_01G2gn9rNFmDtaWTBsMovrLH)_


dvejsada:
@claude review


---

PR #139, opened 2026-09-19
word, excel: give both tools a warnings channel (#114)

### LINKED ISSUE #114: Word and Excel tools drop content silently; give them a warnings channel
Found while writing the developer docs for #111.

The PowerPoint tool returns a `warnings` list alongside the file so the calling model can correct its next call (`PowerpointPresentation.warnings`, wrapped by `main.py` into `{"file", "slide_count", "warnings"}`). The Word and Excel tools have no equivalent. Each of the following produces a success response and only a log line:

**Word (`docx_tools`)**
- `markdown_processor.process_markdown_block()` catches every exception, logs at ERROR, and skips one line. The block is missing from the document.
- `block_elements.add_image_to_doc()` writes a paragraph `[Image could not be loaded: <url>]` on failure.
- `style_map.apply_style()` falls back to `Normal` when a mapped or directive style is missing from the template.
- `add_table_to_doc()` logs and skips a cell it cannot populate.

**Excel (`xlsx_tools`)**
- `parser.walk_markdown_lines()` drops any line that is not a heading, sheet marker, directive or table row.
- `helpers._resolve_row()` resolves a reference to a non-existent table (`T9`) against the current row and logs.
- `helpers._warn_unknown_sheet()` emits a cross-sheet reference to a sheet that does not exist (Excel shows `#REF!`).
- `helpers._write_formula()` stores an over-length formula as text.
- `helpers._ensure_unique_table_headers()` renames headers.
- `styles.parse_styles_directive()` drops malformed entries and oversized ranges.
- `_build_workbook()` warns on colliding sheet names.
- `circular_refs.detect_circular_references()` finds cycles that Excel will resolve to 0.
- Per-cell exceptions in `add_table_to_sheet()` leave the cell empty.

**Proposal.** Mirror the PowerPoint shape:

1. Thread a `warnings: list[str]` through `process_markdown_content()` / `_build_workbook()` (a small context object rather than a global, since builds run concurrently on worker threads).
2. Have each of the sites above append a one-line, caller-actionable message (`"line 42: table could not be rendered (…)"`, `"formula in C7 references T9 but the sheet has 3 tables"`).
3. Return `(BytesIO, warnings)` from `_markdown_to_word_buffer()` and `_markdown_to_excel_buffer()`, and have the handlers in `main.py` wrap the result exactly as the PowerPoint handler does, keeping a bare string when there is nothing to report so existing clients see no change.
4. Update the tool descriptions to mention the `warnings` field, and `docs/markdown-reference.md`.

`AGENTS.md` already states the rule: anything else that grows a warnings channel follows the `(BytesIO, warnings)` shape.

dvejsada:
Closes #114.

## The problem

Both tools returned a success response however much of the caller's input had gone missing on the way.

**Word** skipped a block whose rendering raised, wrote `[Image could not be loaded: …]` for an image that would not download, and fell back to `Normal` for a style the template does not define. **Excel** dropped every line that is not a heading, sheet marker, directive or table row; resolved a formula naming a table or sheet that does not exist against the current row instead; and stored an over-length formula as text. Each was a log line in a response that said "created successfully" — and the caller, usually a model that could have fixed its markdown, never saw it.

PowerPoint has reported this since #122. This gives Word and Excel the same channel.

## What changed

**`warning_channel.py`** — a new shared root module: the severity vocabulary (`error` = not in the file, `warning` = there but not as asked, `info` = a substitution the caller will not mind), the `DocumentWarning` record, and the per-build `WarningChannel`.

The channel is passed down the call chain as an **argument, never module state** — builds run concurrently on `run_blocking` worker threads. Two properties make it safe to call from inside a loop over every cell or line: it de-duplicates identical `(code, message, location)` entries, so a template missing `List Number` warns once rather than once per list item; and it caps what it carries (50), appending one `warnings_truncated` record, so a page of prose fed to the Excel tool does not return hundreds of warnings and eat the caller's context.

**`docx_tools/warnings.py`** and **`xlsx_tools/warnings.py`** own their codes and the one severity per code. `pptx_tools/warnings.py` now takes the severity constants from the shared module so `error` means the same thing in every tool, and keeps its own `SlideWarning` record: a slide index is not a line or a cell, and that shape is already published.

**Entry points** — `_markdown_to_word_buffer()` and `_markdown_to_excel_buffer()` return `(BytesIO, warnings)`, as `_create_presentation_buffer()` does. `main.py` attaches them with a new `_with_warnings()` helper, which the PowerPoint handler now shares. The public sync wrappers (`markdown_to_word`, `markdown_to_excel`) drop them, as `create_presentation` does.

**Nothing to report → nothing changes.** A clean build returns exactly what it returned before: a bare URL string, or the LibreChat artifact dict. Existing clients see no difference.

### Where a warning comes from

Word locates a warning by the 1-based source line; Excel by `sheet` plus `cell`, or `sheet` plus source `line` while the problem is still markdown.

| Word | | Excel | |
|---|---|---|---|
| `block_failed` | error | `line_dropped` | error |
| `table_failed` | error | `table_incomplete` | error |
| `table_cell_failed` | error | `cell_failed` | error |
| `image_failed` | error | `table_reference_missing` | error |
| `table_not_recognised` | warning | `sheet_reference_missing` | error |
| `table_separator_missing` | warning | `formula_unresolved` | error |
| `style_missing` | warning | `formula_too_long` | error |
| `style_fallback_missing` | warning | `circular_reference` | error |
| `widths_invalid` | warning | `table_separator_missing` | warning |
| | | `sheet_name_collision` / `sheet_name_invalid` | warning |
| | | `header_renamed` | warning |
| | | `style_entry_invalid` / `style_range_too_large` / `style_failed` | warning |

`table_separator_missing` is the one code both tools share, because the outcome is identical: a table parsed without a `|---|---|` row under its first row, so that row was made the header. Where the outcome differs the codes do too — pipe markup that is not a table at all is `table_not_recognised` (warning) in Word, where it falls through to the paragraph branch and survives as prose, and `table_incomplete` (error) in Excel, where a worksheet has nowhere to put it.

```json
{"file": "https://…/workbook.xlsx",
 "warnings": [
   {"code": "table_reference_missing", "severity": "error",
    "sheet": "Summary", "cell": "C7",
    "message": "the formula references T9, but the target sheet has no such table (it has: T1, T2); the reference fell back to the current row and almost certainly points at the wrong cell."}
 ]}
```

## What review added

Six review passes ran over this branch; every finding was verified against the code before acting, and each fix has a regression test confirmed to fail on the preceding commit. Four of the six were gaps in `parse_table()`'s reporting, one was a defect in an earlier commit of this PR, and one was a false positive in a warning the PR itself adds:

| Commit | Finding |
|---|---|
| a0efec8 | Pipe lines `parse_table()` consumes without producing a table never reached the dropped-line report — gone from the workbook, nothing said. New `table_incomplete`. |
| 9529d0c | Word's equivalent path renders the line as prose. Nothing lost, but the caller asked for a table: `table_not_recognised`, warning. |
| f2b2d46 | `parse_table()` never *required* the separator row, so a table written without one parsed with its first row silently made the header — and `T1.B[0]` counts from the row *after* the header, so formulas pointed one row off. New `table_separator_missing`. |
| 4028d3a | **Defect in f2b2d46**: `has_separator` answered "did I see a separator anywhere", so a trailing or leading `\|---\|---\|` suppressed the very warning it was added to raise. It now means "directly under the first row". Word gained the same flag and code. |
| 3374507 | Word's separator check skipped empty cells before testing them, and `all()` of nothing is true — so `\|  \|  \|` passed as a separator and the caller's blank row **disappeared**. |
| 29e371a | Excel demanded three dashes where CommonMark and Word accept one, so `\|--\|--\|` was written into the sheet as literal dashes **and** reported `table_separator_missing` against a table that had one. |

**Two commits change parsing, not just reporting**, and they are called out because the rest of the PR deliberately does not: 3374507 (a blank row is kept as data instead of being swallowed) and 29e371a (a short-dash separator is consumed instead of becoming a data row). Both make Word and Excel agree, and both replace a silent drop with something visible and reported. The full suite is unchanged by either.

PowerPoint's own `_SEPARATOR_CELL_RE` still requires three dashes and is untouched: its tables come from structured slide data, not this markdown path.

## Tests

Three new files; `ruff check .` clean and the full suite (1966 tests) green.

- `tests/test_warning_channel.py` — the shared contract: severity from the code, the published dict shape, the location spread in rather than nested, de-duplication, the cap and its notice, two builds never sharing a channel.
- `tests/test_docx_warnings.py` — every Word code at its own site, the source line, a missing style reported once across four list items, and the tool boundary: warnings reach the caller as objects, a clean document still returns a bare URL.
- `tests/test_xlsx_warnings.py` — every Excel code, sheet/cell/line locations, a wall of prose capped rather than returned in full, an oversized style range reported once (not twice), and the same two tool-boundary tests.

Each tool has a `test_every_code_has_a_severity`, mirroring PowerPoint's, so a new code cannot skip the table. Where a message makes a claim about behaviour, a test checks the claim rather than the message: `=T1.B[0]` really does resolve to `=B2`, and the blank row really is still in the Word table — which is what makes `warning` rather than `error` the honest severity there.

## Docs

Updated alongside the code in every commit: the `warnings` field in both tool descriptions and in `docs/markdown-reference.md`; a "The warnings channel" section with the code table in the Word and Excel development pages; `warning_channel.py` in `shared-modules.md` and both root module maps; the response shape and the new pipeline stage in `architecture.md`; the rule in `AGENTS.md`, `adding-a-tool.md` and `testing.md`. The "no warnings channel" known limitation is gone from both tool pages, and #114 now reads there as the invariant to keep rather than the gap to close.

## One thing worth a reviewer's opinion

`line_dropped` is an `error` under the rubric ("something the caller asked for is not in the file"), which means ordinary prose in Excel markdown comes back as an error-severity warning. That is literally true and the cap keeps it from being noisy, but if you would rather it were a `warning`, it is one line in `xlsx_tools/warnings.py`. Review argued for keeping it as `error` — downgrading would make a real loss of content easier to overlook — and I've left it.


dvejsada:
@claude review

Please review this PR. It adds a warnings channel to the Word and Excel tools (#114), mirroring PowerPoint's: a shared `warning_channel.py` (record + per-build collector, de-duplicating and capped), per-package code/severity tables, `(BytesIO, warnings)` from both buffer functions, and a `_with_warnings()` helper in `main.py` that the PowerPoint handler now shares too.

Worth particular attention:

- **Threading the channel.** It is an argument everywhere, never module state, because builds run concurrently on `run_blocking` worker threads. Please check I did not miss a path where a warning is raised without a channel, or where one build's channel could reach another.
- **Backwards compatibility.** A build with nothing to report must return exactly what it returned before (a bare URL string, or the LibreChat artifact dict). The `_with_warnings()` refactor also changed the PowerPoint handler's code — the behaviour should be identical.
- **Severity choices** in `docx_tools/warnings.py` and `xlsx_tools/warnings.py`, especially `line_dropped` as an `error`: it is literally "content the caller wrote that is not in the file", but it fires on ordinary prose in Excel markdown.
- **Sites I may have missed** — anything else in `docx_tools`/`xlsx_tools` that skips, substitutes or degrades silently and should be on the channel.

---

dvejsada:
Thanks — the gap is real, and I've fixed it in a0efec8.

**Confirmed against the code.** `parse_table()` ([`xlsx_tools/helpers.py:229`](https://github.com/ForLegalAI/mcp-ms-office-documents/blob/a0efec8/xlsx_tools/helpers.py#L229)) consumes the whole run of `|` lines and returns `(None, i)` when it cannot build a table from them, so those lines are already past the point where the `else` branch could report them. Two ways in, both silent before this:

- a row with no separator (`| Item | Qty |` alone) — `len(table_lines) < 2`;
- separator rows with no header (`|---|---|` twice) — every line is a separator, so `table_data` comes back as an empty `TableData`, which is falsy and takes the same path.

**The fix.** The branch that consumes the lines now owns reporting them, under its own code rather than falling through to `line_dropped`. The two say different things to the caller: `line_dropped` means "this belongs in a heading or a cell", while here the content *is* tabular and the fix is a specific one — add the separator row. So `table_incomplete` (severity `error`, same rubric — content the caller wrote that is not in the workbook):

> `'| Item | Qty |'` starts something that looks like a table but is not one, so it is not in the workbook. A table needs a header row, a separator row (|---|---|) and at least one data row.

Two regression tests, one per shape, in `TestContentThatDidNotArrive`. I verified they fail on the previous commit and pass on this one — not just that they pass now. Docs updated in the same commit: the code table in `docs/development/tools/excel.md`, the pipeline line, the code list in `docs/markdown-reference.md`, and a second "rule for the sites" noting that a branch which consumes lines owns reporting them.

I also checked the equivalent path in Word, since it has the same shape: `process_markdown_block()` calls `parse_table()` and, on a falsy result, does **not** return — it falls through the rest of the dispatch chain to the paragraph fallback, so the line is rendered as prose rather than lost. No warning needed there.

**On `line_dropped` staying `error`** — agreed, and for the reason you give: the rubric is about the document, and downgrading it would make a real loss of content easier to overlook. Left as is.

`ruff check .` clean; full suite green at 1948 tests (3 new).

---

dvejsada:
@claude review

Second pass, please — the previous round found one real gap (pipe lines that `parse_table()` consumes without producing a table were dropped with nothing reported) and a0efec8 fixes it with a new `table_incomplete` code, two regression tests, and the docs.

Please re-check that fix specifically:

- the new branch in `xlsx_tools/parser.py` — does it cover every way `parse_table()` can come back empty, and is `table_start + 1` the right line to point at?
- whether a separate code was the right call versus reusing `line_dropped`;
- anything else in `docx_tools`/`xlsx_tools` where a branch consumes or swallows caller input before the reporting site can see it — that was the shape of the last finding, so there may be more of it.

---

dvejsada:
On the non-blocking Word note (unrecognised table markup renders as prose with no warning) — I've taken it, in 9529d0c. You're right that it is not a bug, but it is the rule this PR adds applied to a site the PR already changes, so leaving it out would make the two tools disagree about the same input.

`docx_tools/block_elements.py:71` returns `None` when fewer than two pipe lines are collected, and the dispatcher does not return on a falsy result — the line falls through to the paragraph branch and is written with its pipes intact. New code `table_not_recognised`, severity **warning**, not error: the text is still in the document.

That difference is deliberate and now stated in `docs/development/tools/word.md` — severity is about the document, so the same input rates a `warning` in Word, where the line survives as prose, and an `error` (`table_incomplete`) in Excel, where a worksheet has nowhere to put it. Different code names too, since they tell the caller different things about what became of their content.

Three tests: the warning and its line, that the text really is still in the document (so the severity is honest), and that a well-formed table reports nothing.

`ruff check .` clean; suite green at 1951.

---

dvejsada:
Right on both counts — the third shape was uncovered, and it is the most likely one. Fixed in f2b2d46.

**Verified first.** `parse_table(['| Item | Qty |', '| A | 1 |'], 0)` returns a `TableData` of two rows with `col_alignments == []` and `bool(...) is True`, so it becomes a real `TableEvent` and neither new code fires. The consequence is worse than cosmetic: `base_xlsx_tool.py` treats `table_data[0]` as the header unconditionally, so the first row is styled as one, dropped from the data, **and** counted out of every table-relative reference — `T1.B[0]` means the first row *after* the header. A caller who meant both rows as data gets formulas pointing one row off, silently.

**`table_separator_missing`, severity `warning`** — not `error`, and behaviour unchanged. Nothing is missing from the workbook and the lenient reading is usually what the caller meant; what is worth saying is that the tool decided it for them, and what that decision costs:

> the table starting here has no separator row (|---|---|), so its first row was used as the header and the rest as data. Add the separator to say which row is the header — references like T1.B[0] count from the first row after it.

**On plumbing:** the fact could be inferred in `parser.py` from `col_alignments` being empty, but that is an accident of how alignments are parsed, not a statement about the markdown. `TableData` now carries `has_separator` explicitly, defaulting to `True` — only markdown that was really parsed can say otherwise, so no other construction of a `TableData` can produce a spurious warning.

Three tests: the warning and its line; a test asserting the behaviour the message describes (`A1` is `Item`, `A2` is `A`, and `=T1.B[0]` resolves to `=B2`), so the claim in the message is checked rather than just asserted; and a well-formed table reporting nothing. I confirmed the first fails on 9529d0c and passes here. The behaviour test passes on both, deliberately — it documents pre-existing behaviour this commit does not change.

Also confirmed from your trace of `table_start + 1`: it is the first line of the consumed run, same 1-based convention as `line_dropped`, and the new warning reuses it.

Docs: the code table and a paragraph in `docs/development/tools/excel.md` explaining why `parse_table()` is lenient and what carries the fact out of the parse; the code list in `docs/markdown-reference.md`.

`ruff check .` clean; suite green at 1954.

---

dvejsada:
@claude review

Third pass, please — on the current head (f2b2d46). Since the last review two things landed:

- **f2b2d46** — the third `parse_table()` shape you found: a table with no separator row at all parses as real, with its first row silently made the header. New `table_separator_missing` code at `warning` severity (nothing is lost, and the lenient reading is usually right — what is reported is that the tool decided it, and that `T1.B[0]` counts from the row after the header). `TableData.has_separator` carries the fact out of the parse.
- **9529d0c** — the non-blocking Word note from your first pass, taken for parity: `table_not_recognised`, severity `warning` there because the line falls through to the paragraph branch and survives as prose.

Please check:

- whether `parse_table()` now has any *fourth* shape that still slips through, in either tool;
- the same-input/different-severity split between the two tools (`warning` in Word, `error` in Excel) — is that a distinction a caller can actually use, or is it confusing?
- the `TableData.has_separator` default of `True` — safe, or does it hide a case?
- anything in the three commits' tests that asserts the implementation rather than the behaviour.

---

dvejsada:
Both taken, in 4028d3a. The first is a flaw in the previous commit, so I'd have wanted it whatever its blocking status.

**The fourth shape.** Verified before fixing — `parse_table(['| A | 1 |', '| B | 2 |', '|---|---|'], 0)` returned `has_separator: True`, and so did the leading-separator form. The flag was answering "did I see a separator row anywhere in this run", which is not what it was added to mean, so it suppressed `table_separator_missing` in exactly the case the warning exists for: the first row silently made the header. It now means *the separator was directly under the first row* — the only position markdown gives it meaning. Parsing is unchanged: a separator found elsewhere is still skipped, it just no longer counts as one.

```
well formed  -> has_separator: True
trailing sep -> has_separator: False
leading sep  -> has_separator: False
no separator -> has_separator: False
```

**Word's counterpart.** You're right that Word's `parse_table()` had the same leniency, and that the outcome is the tool picking the header row. It now returns the flag as a fourth value — one caller, `markdown_processor.py`, unpacks it — and reports the same code.

**On naming**, I went with the same code in both tools rather than a separate `table_header_assumed`, which I think is the more useful answer to your severity-split question from the third pass. The two are different situations:

- Same outcome in both tools → same code, same severity. A table whose header was chosen for the caller is `table_separator_missing`/`warning` in Word and Excel alike; only the *consequence* differs (Word has no `T1.B[0]` offsets to shift), and consequence is not what the code names.
- Different outcome → different code. `table_incomplete` (error, Excel) vs `table_not_recognised` (warning, Word) stay separate precisely because the content survives in one tool and is gone in the other, which is what a caller branching on severity needs to tell apart.

So the split you asked about is not "the same thing rated two ways" — it is two different fates for the input, and where the fate is identical the report now is too. Both are documented in the Word and Excel pages.

**On `has_separator` defaulting to `True`** — it is only ever passed explicitly by `parse_table()`, the sole place a `TableData` is built from markdown. The default exists so a `TableData` constructed any other way (a test fixture, future code) cannot produce a spurious warning about markdown it never saw.

**On tests asserting implementation** — the closest was `test_the_warning_describes_what_actually_happened`, which asserts `=T1.B[0]` resolves to `=B2`; that is the behaviour the message claims, so I kept it. This round adds `test_the_table_is_still_built`, which checks every row the caller wrote is in the Word table — the thing that makes `warning` rather than `error` the honest severity. Four new tests total; I confirmed all four fail on f2b2d46 and pass here.

`ruff check .` clean; suite green at 1959.

---

dvejsada:
@claude review

Fourth pass, please — on the current head (4028d3a), which fixes both third-pass findings: `has_separator` now means "directly under the first row" in both tools, and Word reports `table_separator_missing` too.

Specifically:

- is there a *fifth* shape, or is `parse_table()`'s reporting now complete in both tools? If you find one, please say whether it is genuinely reachable from markdown a caller would write, rather than only constructible;
- the naming decision: same code (`table_separator_missing`) where the outcome matches across tools, different codes (`table_incomplete` / `table_not_recognised`) where the content survives in one and not the other — does that hold up from a caller's point of view?
- whether the flag's meaning change silently altered any existing parsing behaviour, rather than only what gets reported.

If the remaining items are all narrower than what a caller would plausibly write, I would rather hear that plainly than keep widening the PR — I want to close #114 on a complete channel, not an infinitely deep one.

---

dvejsada:
Taken, in 3374507 — and it turned out to be more than a suppressed warning, which is why I did not leave it as a follow-up.

**What I found on verifying it.** Your trace of the vacuous `all()` is exactly right. But separator rows are *dropped* rather than rendered, so the blank row does not just fail to warn — it disappears:

```
parse_table(['| Item | Qty |', '|  |  |', '| A | 1 |'], 0)
  before → rows: [['Item','Qty'], ['A','1']]                 has_separator: True
  after  → rows: [['Item','Qty'], ['',''], ['A','1']]        has_separator: False
```

Three rows in, two out. A caller who wrote a blank spacer row lost it with nothing reported — silent content loss, which is the thing this issue is about, rather than only a missing report. So I took it on that basis rather than on where it came from; pre-existing and non-blocking are both true and neither changes that it loses content.

One change fixes both: an empty cell now fails the match, so the row is data, it survives, and the table is correctly reported as having no separator. Excel's `_is_separator_row()` already checked every cell, so this is the two tools agreeing rather than a new rule.

**Flagging one thing, since it is the exception to how the rest of this PR behaved.** This commit changes *parsing*, not just reporting: a malformed separator with a blank column (`| --- |  | --- |`) now renders as a data row instead of vanishing. I think visible-and-reported beats silently-dropped, and it is the same principle as the blank row, but it is a behaviour change and you were right to ask about that category. The full suite is unaffected by it — 1961 green, including the Word table tests that predate this PR — and two new tests pin both directions: a blank row survives and warns, a real `|:--|--:|` separator is still consumed and reports nothing.

**On your other two answers** — agreed on both, and thank you for checking the naming from the caller's side rather than just the code's. Noted too that you found the header-count, duplicate-separator and separator-only-body cases render what was written and so are not this channel's business; that matches what I'd concluded, and it is useful to have it confirmed independently rather than assumed.

`ruff check .` clean; suite green at 1961; CI was green on 4028d3a.

---

dvejsada:
@claude review

Fifth pass, please — on the current head (3374507), which takes the blank-cell separator finding. It turned out to lose content, not just suppress a warning: separator rows are dropped rather than rendered, so `|  |  |` under the header made the caller's blank row disappear. One change fixes both, and Word now matches Excel's `_is_separator_row()`.

Two things specifically:

- **The behaviour change.** This is the one commit in the PR that alters parsing rather than only reporting: `| --- |  | --- |` now renders as a data row instead of being consumed. Is that the right call, and does anything else depend on the old vacuous match that the suite would not catch?
- **Whether the channel is now complete** for the shapes a caller would plausibly write. Your last pass said `parse_table()`'s reporting was complete apart from this one; if that still holds with this in, please say so plainly — I would like to close #114 on this rather than keep going if what is left is only constructible input.

---

dvejsada:
Thanks — and I've taken the asymmetry you flagged in passing, in 29e371a, because checking it turned up a cost you'd ruled out.

**One correction to the finding.** You wrote that the dash-count difference is "not a content-loss issue (a too-short dash run just falls through to `table_separator_missing` on the Word side and behaves the same on Excel's)". It doesn't behave the same. `|--|--|` is a valid separator by CommonMark and by Word's `-+` check, but Excel's `-{3,}` rejects it, so Excel:

1. writes the caller's separator into the sheet as a data row of literal `--`, and
2. reports `table_separator_missing` against a table that plainly has one.

```
|-|-|       -> warnings: ['table_separator_missing'] | A2 = '--'      (before)
|-|-|       -> warnings: none                        | A2 = 1         (after)
```

The second part is what decided it: that is the channel this PR adds telling the caller something untrue about their own markdown. A false positive in a new warning is this PR's to fix, not a follow-up.

**Same caveat as last time, and it is the second of two.** This changes parsing as well as reporting — a short-dash separator is now consumed rather than becoming a data row. It is the correct reading and the one Word already gave it, so the two tools now agree, but I would rather name it than let it pass as routine. Five new cases pin `|-|-|`, `|--|--|`, `|---|---|`, `|:-:|--:|` and `|:---|---:|`: nothing reported, and the row below the separator is the first data row. Suite unchanged at 1966.

PowerPoint's own `_SEPARATOR_CELL_RE` still requires three dashes. I left it alone deliberately — its tables come from structured slide data, not this markdown path, so it is neither part of this channel nor covered by the reasoning above. Worth its own look sometime; not this PR.

**On your verdict** — noted, and the shape-space walk is a better answer than I'd have written for myself. I'm treating the channel as complete on the five shapes you list.

---

dvejsada:
@claude review

Sixth and, I expect, final pass — on 29e371a, which is one commit on top of the head you just cleared. It takes the dash-count asymmetry you flagged as a non-blocking aside, because on checking it produced a false `table_separator_missing` against a table that had a valid CommonMark separator, and wrote that separator into the sheet as a data row of literal dashes.

Please confirm just that commit:

- the regex change (`-{3,}` → `-+`) is right, and nothing in the Excel path depended on the stricter form — in particular anything that treats a dash row as data;
- the parsing change is acceptable on the same grounds as the Word one, or if not, say so;
- leaving PowerPoint's `_SEPARATOR_CELL_RE` at three dashes is the right scope call, given its tables come from structured slide data rather than this markdown path.

You've already said the channel is complete on the five shapes; I'm not asking you to re-open that. If this commit is sound I'll take the PR as done and leave it to its reviewers.

---

dvejsada:
Taken rather than deferred, in 8c8ebe6 — because you identified the part that makes it mine: **29e371a made it reachable.**

You framed it as widening the trigger on a pre-existing gap. That understates it from the caller's side: before that commit a lone `-` was not a separator in Excel, so `| - | - |` survived; after it, the same row disappears. For that input it is a regression my commit introduced, two commits ago, and "pre-existing logic" is not much comfort to whoever loses the row.

```
| Item | Qty |     rows written before 29e371a: Item/Qty, A/1, B/2
|---|---|                    after 29e371a: Item/Qty, A/1, B/2   ← '-' row gone
| A | 1 |                         after 8c8ebe6: Item/Qty, A/1, -/-, B/2
| - | - |
| B | 2 |
```

**The fix is the rule you proposed**, in both tools: markdown has exactly one separator and it sits directly under the header, so position decides and shape only confirms it. Row 1 of the run may be the separator; every other row is content, whatever it looks like. That also retires the "matches anywhere, drops unconditionally" asymmetry you named — `separator_in_place` no longer gates only the *reporting* while the *drop* happens regardless.

**One interaction worth naming, since it needed handling rather than falling out.** A run of nothing but separator rows (`|---|---|` twice) relied on the old drop-anywhere behaviour to end up with zero rows and be reported `table_incomplete` — the second-pass finding. Under a naive `idx == 1` rule that becomes a table whose header reads `---`, silently. So there is now an explicit early return for that case, keeping the second-pass behaviour deliberately rather than regressing it to make this commit simpler. Two regression tests pin the new behaviour (confirmed failing on 29e371a) and the existing `test_separator_rows_with_no_header_are_reported` still passes, which is what guards the interaction.

This is the third commit that changes parsing rather than only reporting, and by the same reasoning as the other two: content the caller wrote stops disappearing. Suite green at 1968.

On your other two answers — agreed, and thank you for tracing all five inputs by hand rather than taking the parametrized test's word for it. PowerPoint stays untouched.

---

dvejsada:
@claude review

Seventh pass — on 8c8ebe6, which takes the dash-row finding rather than deferring it, since 29e371a is what made it reachable.

Narrow scope, please, just this commit:

- the `idx == 1` rule in both `parse_table()`s — does it hold for every run shape, and did I miss a case that used to depend on the drop-anywhere behaviour?
- the explicit early return for an all-separator run, which exists to keep `table_incomplete` working under the new rule — is that the right way to preserve it, or does it paper over something?
- whether three parsing changes across this PR now leave the two tools genuinely consistent, or whether they have drifted somewhere I have not looked.

This has been a productive run — six passes, seven findings, every one real, and two of them defects in my own earlier commits here. If this commit is sound I'll stop and leave the PR to its human reviewers rather than keep asking.

---

dvejsada:
Correct on all three, and the third one is the important one: they had drifted, and I drifted them. Fixed in d098105.

**Confirmed exactly as you traced it**, including the generalisation past two rows:

```
before d098105:
  |---|---| ×2  -> table: [['---', '---']]                  warnings: none
  |---|---| ×3  -> table: [['---', '---'], ['---', '---']]   warnings: none
after:
  |---|---| ×2  -> tables: 0, paragraphs: ['|---|---|', '|---|---|'],  warnings: 2× table_not_recognised
  |---|---| ×3  -> tables: 0, paragraphs: 3× '|---|---|',              warnings: 3× table_not_recognised
```

**And it was a regression, not just an omission.** Under the old shape-only matching, every separator-looking line was consumed, `table_data` came out empty, and the lines fell through to prose. 8c8ebe6's `idx == 1` rule stopped consuming them, so one became the separator and the rest became content — which means the commit that closed this class of bug in Excel reopened it in Word. That is the fourth finding in this chain caused by one of my own commits rather than by the original code, and the pattern in all four is the same: a change that was right where I applied it, and unexamined one layer over.

**Word takes the not-a-table path**, as you suggested, rather than Excel's `table_incomplete`: the lines are reported as `table_not_recognised` and written as prose, so the caller keeps what they wrote. Excel reports `table_incomplete` for the same input because a worksheet has nowhere to put it. Codes differ where the outcome does, as elsewhere here.

Two further things, since "they cannot drift" is worth more than "they do not currently differ":

- The shape test now lives in `_is_separator_line()`, used by both the guard and the loop, so those two can no longer disagree about what a separator row looks like. Position stays with the caller of it, which is the one rule that differs between the guard (any row) and the loop (row 1).
- Word had **no test for this shape at all** — which is exactly why the drift went unnoticed when I wrote 8c8ebe6, since Excel's `test_separator_rows_with_no_header_are_reported` passed and I took that as covering the behaviour. It now has one, parametrised over runs of two and three, confirmed failing on 8c8ebe6.

Suite green at 1970.

---

dvejsada:
@claude review

Eighth pass — on d098105, which ports the all-separator guard to Word and factors the shape test into `_is_separator_line()` so the guard and the loop cannot disagree.

I said last time I'd stop asking if the commit was sound, and then the commit wasn't. So, narrowly:

- **the two `parse_table()`s side by side.** Not "do they agree on the inputs we have tests for", but: is there any run shape where they still diverge without a deliberate reason? Four of the eight findings so far have been my own commits regressing something one layer over from where I was looking, and this is the question that keeps catching it.
- whether routing Word's all-separator run to `table_not_recognised` + prose is right, versus Excel's `table_incomplete`.
- whether `_is_separator_line()` is now the single shape test in Word, with no second spelling of it left anywhere.

If this one is sound I'll stop for real and hand the PR to its human reviewers.

---

dvejsada:
Thanks — and thank you for flagging that you couldn't run `pytest`/`ruff` this pass. That's the right thing to declare rather than let a hand-trace read as an executed check, so I re-ran the three claims that carried weight:

```
1. single shape test in Word
   _SEPARATOR_RE / ':?-+:?' appear only inside _is_separator_line() (lines 40, 53, 56)
   call sites: 106 (guard), 119 (loop) — both via the function.  ✓

2. trailing-pipe divergence pre-existing
   git show 3333d01:xlsx_tools/helpers.py      -> 220: if not line.endswith('|'):
   git show 3333d01:docx_tools/block_elements.py -> 68: startswith('|') and endswith('|')
   both present before this PR's first commit.  ✓

3. '| Item | Qty' / '| A | 1' (no trailing pipes), today
   excel: ['table_separator_missing']
   word : no warning, paragraphs: ['| Item | Qty', '| A | 1']   ✓ matches your trace
```

`ruff check .` clean and the suite green at 1970 on this head, which is the part your pass couldn't confirm.

**On the trailing-pipe divergence — agreed, leaving it.** It is real and it is a divergence, but it is not this channel's business and not a content-fidelity bug in either direction: Excel pulls the line into the table machinery and reports it, Word never enters the table branch and renders the text as visible prose. Nothing is lost or silently reinterpreted on either side, which is the line this PR has been drawing throughout. It predates the whole chain, so fixing it would mean changing which lines are tables — a separate question from what gets reported when they are. Worth its own look; not this PR.

**Taking this as done.** Eight passes, nine findings, every one verified against the code before acting and every fix carrying a regression test confirmed to fail on its preceding commit. Four of the nine were defects in my own commits in this PR rather than in the original code, which is the part worth recording: the review chain was catching my regressions, not just the pre-existing gaps, and the question that caught them each time was "do the two implementations still agree off the tested path", not "do the tests pass".

Handing it to human reviewers now. CI green on every head; no conflict with `master`.

---

PR #140, opened 2026-09-19
xlsx: keep a percent column's precision in its formula cells (#126)

### LINKED ISSUE #126: xlsx: a formula in a percent column gets a flat 0% format
Different tool, same repo. `_number_format_for_type()` returns `'0%'` for
`percent`, and the formula branch applies the column's number format verbatim.
So in a column declared `types: percent`, a computed growth rate of 4.3% renders
as `4%` — the precision is silently lost, and only for formula cells, since
literal values carry their source precision through
`_percent_format_for()`.

Workaround is to declare `number:0.0%` instead, which is not discoverable.

**Suggested fix:** let a `percent` column declare its precision
(`percent:0.0%`), or default formula cells in a percent column to the precision
used by the literal values elsewhere in that column.

dvejsada:
Closes #126.

## The problem

A literal percent takes its format from its own source text, so `4.3%` keeps its decimal through `_percent_format_for()`. A formula has no source text, and `_number_format_for_type()` returned a flat `0%` for a `percent` column — so a computed growth rate rendered as **`4%` sitting directly beside literals that had kept theirs**.

The loss was silent, hit only the computed cells, and the documented workaround was to abandon `percent` for `number:0.0%` — which nothing in the tool description or the reference mentioned.

```
<!-- types: text, percent -->
| Metric | Rate |
|--------|------|
| A      | 4.3% |     ->  0.0%   (kept its precision)
| B      | 5.1% |     ->  0.0%
| Mean   | =AVERAGE(B2:B3) |  ->  0%   before, 0.0% after
```

## The fix

Both halves the issue proposed, because they answer different needs.

**A bare `percent` column learns from its own literals.** The common case then needs nothing from the caller. The *widest* precision in the column wins: showing more decimals than a neighbour is a cosmetic surprise, showing fewer hides part of the number. A column with no literal percent to learn from keeps the flat `0%`.

**`percent:<format>` declares it outright**, which `number:`, `date:` and `currency:` already allowed, and then applies to every cell in the column — literal and computed alike, since uniformity is the point of declaring it.

Values are untouched throughout: this is display precision only. `4.25%` stays `0.0425` in the cell whatever format it is shown with.

## What counts as a literal

The learner and the column coercion must agree on this, and twice they did not — the whole of the review on this PR was about that one seam. They now run the same two operations on the same string (`rstrip('%')`, then a bare `float()`), so the answer is whatever `_apply_column_type()` would actually do with the cell:

| literal | the column coerces it to | teaches the formula cells |
|---|---|---|
| `4.3%` | `0.043`, shown `0.0%` | `0.0%` |
| `50.5` (no sign — the `%` is optional) | `0.505`, shown `0.0%` | `0.0%` |
| `007` | `0.07`, shown `0%` | `0%` |
| `1,234%` | nothing — stays text, `General` | nothing |
| `2024-01-15`, `yes`, `N/A` | nothing — stays text, `General` | nothing |
| `=B2*1.5%` | a formula, not a literal | nothing |

The rule is that what the column *does* with a value decides whether it teaches its precision — not the sign it was written with, and not whether its comma parses.

## Two things the issue didn't mention

Both turned up while implementing, and the explicit form would not have worked without the first:

- **`_apply_column_type()` matched the bare word `percent` exactly**, so a suffixed `percent:0.00%` fell straight through to generic cell resolution — the value was still coerced, but by the untyped path, and the declared format was ignored entirely.
- **The column-width estimate had the same exact match**, so `percent:<format>` fell into the `else` branch and estimated from raw text length rather than the format.

`_parse_types_directive()` already kept `percent:#,##0.0%` in one piece — the comma-rejoining logic covers it because `percent` is a known keyword — and there is now a test pinning that, since it is the kind of thing a later change could quietly break.

## What review added

Three passes, three findings, all real — and two of the three were in this PR's own new code, both the same seam in opposite directions:

| Commit | Finding |
|---|---|
| 25e1c57 | The learner was **more lenient** than the coercion: it replaced the comma before `float()`, so `1,234%` — which the coercion rejects, leaving the cell as text — was setting the format of the computed cells beside it. |
| beb928c | The learner was **stricter**: it required a trailing `%`, which the coercion does not, so a bare `50.5` became a real percent in the sheet and taught nothing. |

After the second, the fix stopped being another patch: both tests are now literally the coercion's own, so the two cannot drift again. The third pass confirmed that structurally rather than by example — same source string, same `startswith('=')`, same `rstrip('%')`, same bare `float()` — which is the only form of "they agree" worth having after a passing suite missed it twice.

One item was raised and deliberately left: the `.replace(',', '.')` inside `_percent_decimals()` is unreachable, since all three callers gate on a bare `float()`. Unreachable is not wrong, and removing it would change a helper's documented contract and its test for no behavioural gain. The consequence it points at — a percent column does not accept `4,3%` at all — is pre-existing, orthogonal to precision, and wants its own issue.

## Tests

Eleven new cases in `tests/test_xlsx_tier2_semantics.py`, which already owns percent precision. Seven fail without their fix; I confirmed that rather than only checking they pass now. The rest pin behaviour that must **not** change:

| Test | Guards |
|---|---|
| `test_formula_takes_the_precision_of_the_columns_literals` | the reported bug |
| `test_the_widest_precision_in_the_column_wins` | mixed `4.3%` / `5.15%` — literals keep their own, the formula takes `0.00%` |
| `test_a_percent_written_without_its_sign_still_teaches` | `50.5` is coerced, so it teaches |
| `test_a_value_the_column_could_not_coerce_teaches_nothing` | `1,234%` stays text, so it does not |
| `test_a_rejected_value_does_not_mask_a_real_one` | `1,234%` + `4.3%` → `0.0%`, not `0.000%` |
| `test_the_declared_format_applies_to_literals_and_formulas` | `percent:0.00%` reaches both |
| `test_the_declared_format_wins_over_the_source_precision` | and the stored value is still `0.0425` |
| `test_whole_percent_column_is_unchanged` | a `50%` column still gets `0%` |
| `test_a_column_with_no_literal_percent_falls_back` | nothing to learn from → flat `0%` |
| `test_a_formula_is_not_mistaken_for_a_literal` | `=B2*1.5%` ends in `%` but offers no precision |
| `test_the_spec_survives_the_comma_in_a_format` | `percent:#,##0.0%` is not split in two |

`ruff check .` clean; full suite green at 1981, and CI green on every head.

## Docs

Updated alongside the code: the `types` option list in the tool description in `main.py` and in `docs/markdown-reference.md`, both now naming `percent:<format>` and the precision rule; "Cell resolution" in `docs/development/tools/excel.md`, which described the literal rule only; and a note in "The three directives" on which specs take a number-format suffix, with `_number_format_for_type()` named as the one place that mapping lives.


dvejsada:
@claude review

Please review this PR. It fixes #126 — a formula in a `types: percent` column got a flat `0%`, so a computed 4.3% rendered as `4%` beside literals that kept their decimals. A bare `percent` column now takes its formula format from the column's own literals (widest precision wins), and `percent:<format>` declares it outright for every cell.

Worth particular attention:

- **The "widest precision wins" choice.** The alternative was most-common, or per-cell. Widest never hides a digit, but it can show a computed value with more decimals than any literal beside it. Is that the right trade, and does it hold for a column mixing `50%` with `4.25%`?
- **`_percent_format_for_column()` deciding what counts as a literal.** It skips cells starting with `=` and anything not ending in `%`, and requires the body to parse as a float. A formula like `=B2*1.5%` ends in `%` — there is a test for it — but I would like a second pair of eyes on what else could be misread as a literal, or wrongly skipped.
- **The explicit form reaching every site.** `_apply_column_type()` and the width estimate both matched the bare word `percent` exactly and ignored a suffixed spec; I changed both to `startswith`. Is there a third site I have missed that still assumes the bare form?
- **Anything that regressed.** This changes displayed formats in columns that previously got `0%`. The suite is green, but the tests may simply not have covered a case.

Same posture as #139, please: verify against the code rather than the description, and say plainly if something is pre-existing or out of scope rather than in this PR's territory.

---

dvejsada:
Clean pass, but your point 2 turned up something real when I checked it — and it is not pre-existing. Fixed in 25e1c57.

**The `1,234%` case is mine.** You judged it "the same comma-decimal-vs-thousands ambiguity that already exists for percent literals outside typed columns; not something this PR introduces or worsens". The first half is right about `_percent_format_for()`. But inside a typed column the two functions disagreed, and only one of them is new:

```
<!-- types: text, percent -->
| A    | 1,234% |      -> cell value '1,234%', format General   (coercion REJECTED it)
| Calc | =B2*2  |      -> 0.000%  before this commit,  0%  after
```

`_apply_column_type()` does a bare `float('1,234')`, which raises — the cell stays text. My `_percent_format_for_column()` replaced the comma first, so it parsed, counted three decimals, and set the format of the computed cells from a value that is not a percent in the finished sheet. Before this PR that formula got `0%`; the learner is what changed it, two commits ago.

**The fix** is to make the learner's acceptance test the same bare `float()`, so it and the coercion agree on what a percent is. The looser parse stays in `_percent_decimals()`, where it belongs — counting the decimals of a value already accepted. Two tests, both confirmed failing on 9ed9f99: the rejected value teaches nothing, and it does not mask a real one beside it (`1,234%` + `4.3%` → `0.0%`, not `0.000%`).

**On the rest** — agreed, and thank you for tracing it against `origin/master...HEAD` rather than my description; the `learned_percent_formats` gate at helpers.py:1132 being a deliberate exact match is exactly the distinction I wanted a second reader on. The `percentile` note I'm leaving: `currency`/`number`/`date` all use `startswith`, so matching them is the consistent choice, and inventing a guard for a spec that does not exist would be the odd thing.

**On what you could not verify** — flagging that was the right call rather than repeating my numbers back. You do not have to take mine either: CI ran `lint-and-test` on 9ed9f99 and it passed, independently of the PR description. It is green on this head too, and the suite is at 1980 now.

---

dvejsada:
@claude review

Second pass, please — on 25e1c57, one commit on top of the head you cleared. It makes `_percent_format_for_column()`'s acceptance test the same bare `float()` the column coercion uses, so a value like `1,234%` that stays text cannot set the format of the computed cells beside it.

Narrowly:

- **Do the learner and `_apply_column_type()` now accept exactly the same set?** Both start from `_strip_markdown_formatting(...)[0].strip()` and `rstrip('%')` as far as I can tell, so the bare `float()` should make them identical — but that is the claim worth checking rather than trusting, since the whole point of the commit is that they had silently diverged.
- **Anything a percent column legitimately coerces that the learner now skips**, i.e. the opposite direction of the bug: a value that does become a percent but no longer teaches its precision.
- Whether keeping the comma-tolerant parse inside `_percent_decimals()` is right, given its other caller is the untyped path where `_strip_thousands_separators()` has already normalised.

Same posture: verify against the code, and say plainly if something is pre-existing rather than this PR's.

---

dvejsada:
Both points check out. Point 2 is fixed in beb928c; point 3 I'm leaving, with reasoning.

**Point 2 — taken, and it is this PR's.** Verified exactly as you traced it:

```
<!-- types: text, percent -->
| A    | 50.5  |   -> 0.505, format 0.0%    (coerced — the % is optional here)
| Calc | =B2*2 |   -> 0%  before,  0.0%  after
```

You called it predating 25e1c57. True of the guard, not of the function: `_percent_format_for_column()` arrived in 9ed9f99, which is this PR's first commit, so both directions of its disagreement with the coercion are mine. That is now twice in two passes — too lenient, then too strict — so I have stopped patching the symptom and made both tests *the coercion's own*, `rstrip('%')` then a bare `float()`, with a docstring saying that is the rule and why. What the column does with a value decides whether it teaches its precision; not the sign it carries, not whether its comma parses. Test confirmed failing on 25e1c57.

Behaviour across the cases now, for the record:

| literal | coerced to | learner |
|---|---|---|
| `4.3%` | 0.043, `0.0%` | teaches `0.0%` |
| `50.5` | 0.505, `0.0%` | teaches `0.0%` |
| `1,234%` | stays text, `General` | teaches nothing |
| `N/A` | stays text, `General` | teaches nothing |

**Point 3 — correct, and my framing was wrong.** I said the looser parse "stays where it belongs"; you are right that all three paths into `_percent_decimals()` now gate on a bare `float()`, which rejects commas outright, so the `.replace(',', '.')` is unreachable from production and only the direct unit test exercises it. I checked the three call sites and agree.

I am leaving it, deliberately rather than by omission. `_percent_format_for()` promises to preserve the precision of source text, and tolerating a European decimal comma is part of honouring that promise for any future caller that does not pre-gate on `float()`. Unreachable is not the same as wrong, and removing it would change a helper's documented contract — with its test — for no behavioural gain in this PR. The thing genuinely worth knowing is the *consequence* you surfaced: a percent column does not accept `4,3%` at all, because the coercion has always used a bare `float()`. That is pre-existing, orthogonal to precision, and belongs in its own issue rather than here.

CI green on this head; suite at 1981.

---

dvejsada:
@claude review

Third pass — on beb928c. It makes `_percent_format_for_column()`'s acceptance test identical to `_apply_column_type()`'s in both directions: `rstrip('%')` then a bare `float()`, no `%`-suffix requirement and no comma handling.

The thing I actually want checked:

- **Are the two tests now genuinely the same predicate, or only the same for the inputs we have thought of?** Twice running, the learner and the coercion disagreed and a test passed anyway. So rather than more examples: is there any input where `_apply_column_type()` coerces a cell to a percent and `_percent_format_for_column()` skips it, or the reverse? If the honest answer is that they cannot diverge because they now run the same two operations, say that — that is the answer I am after.
- Whether removing the `endswith('%')` guard opened anything up in a percent column that should not teach precision: a date, a boolean, a value with leading zeros.
- Point 3 from your last pass: I have left the comma branch in `_percent_decimals()` as unreachable-but-harmless, and my reasoning is in the thread above. Push back if you disagree.

---

dvejsada:
That's the answer I was after — a structural one rather than another example. The side-by-side table of the two paths (same source string, same `startswith('=')`, same `rstrip('%')`, same bare `float()`) is the thing that makes "they cannot diverge" checkable rather than asserted, and it is why I stopped adding cases after the second time they disagreed.

Verified your point 2 claims myself, since you again could not run anything:

```
007          -> literal 0.07, 0%       | formula 0%       (coerced, teaches 0 decimals)
2024-01-15   -> stays text, General    | formula 0.0%     (skipped; learned from the 4.3% beside it)
yes          -> stays text, General    | formula 0.0%     (skipped)
```

Dates and booleans fail `float()` in both functions and are skipped by each, as you said. `007` coerces to `0.07` and teaches `0%` — consistent with itself, and the `%`-optional coercion behind it predates this PR, which I confirmed at 4e7a37e.

`ruff check .` clean; suite green at 1981. CI is green on this head independently of my saying so.

**Taking this as done.** Three passes, three findings, all real — and two of the three were my own code: the learner disagreeing with the column coercion, first by being too lenient and then too strict. Worth recording that the fix that stuck was not a third patch but making the learner run the coercion's own test, so the two cannot drift again. The remaining item (a percent column rejects `4,3%` outright, because the coercion has always used a bare `float()`) is pre-existing, orthogonal to precision, and wants its own issue.

Handing it to human reviewers.

---

PR #141, opened 2026-09-19
cleanups from the documentation pass (#116)

### LINKED ISSUE #116: Cleanups found during the documentation pass
Small things noticed while writing the developer docs for #111. None changes behaviour a user relies on; each is a few lines.

- [ ] **`promo_code_block` special case in `email_tools/dynamic_email_tools.py`.** The generic tool body synthesises a `promo_code_block` context value whenever a template declares `promo_code`. It is a leftover from the first example template; the documented mechanism is a Mustache section (`{{#promo_code}}…{{/promo_code}}`), which `docs/templates.md` now shows. Remove the special case once no deployed template relies on it.
- [ ] **Dead helpers in `pptx_tools/helpers.py`.** `coerce_indent_level()` and `SlideHelpers._add_image_from_url()` have no callers outside tests. Remove them and their tests.
- [ ] **`upload_tools/utils.get_content_type()` matches by substring.** `"pptx" in file_name` means a sanitised name such as `notes.pptx_v2.docx` is uploaded as PowerPoint. Match on the extension after the last dot instead, and share the table with `MIME_TYPES` in `backends/librechat.py`, which duplicates it.
- [x] **`TABLE_BOTTOM_SPACING` is defined in both `xlsx_tools/parser.py` and `xlsx_tools/helpers.py`.** Done in #117: the review pointed out that `parser.py` already imports from `helpers.py` and the reverse import does not exist, so there was never a cycle. `parser.py` now imports the constant.
- [ ] **Static email tool defaults `language` to `cs-CZ`** in both the tool parameter and `_create_eml_buffer()`. Consider an `EMAIL_DEFAULT_LANGUAGE` setting or `en-US`.
- [ ] **`admin/analysis.py` and `admin/preview.py`** are the only modules under `admin/` without a per-tool developer page; a short section in `docs/development/dynamic-templates.md` would close the gap.

dvejsada:
Closes #116.

All five open items on the checklist. Two of them are behaviour changes the issue left to the maintainer, and both were decided explicitly rather than by me.

## `get_content_type()` matched by substring

The real bug of the set. `"pptx" in file_name` made `notes.pptx_v2.docx` a PowerPoint, and any name merely containing `xml` an XML document — and `sanitize_filename()` keeps dots, so such a name is reachable from a caller's `file_name`:

```
notes.pptx_v2.docx          ->  PPTX        before,  DOCX now
quarterly.xlsx.backup.docx  ->  DOCX        (right answer, wrong reason)
myxmlfile                   ->  XML         before,  ValueError now
```

The extension after the last dot decides now, from **one** MIME table that `backends/librechat.py` imports rather than keeping its own copy.

The two copies had already drifted, which is the argument for sharing the table rather than just deduplicating it: `.eml` was `application/octet-stream` on the cloud backends and `message/rfc822` on LibreChat. They now agree on the latter, the accurate type — **the one behaviour change in this section**.

The LibreChat side keeps its guess-don't-raise fallback (`mimetypes`, then `application/octet-stream`) and the traditional side keeps raising: LibreChat takes whatever a caller attaches, while the cloud backends upload only what this server generates, so an unknown type there is a bug worth surfacing at the call site. `get_content_type()` had no tests at all; it has 24 now.

## Dead PowerPoint helpers

`coerce_indent_level()` and `_add_image_from_url()` had no callers. The first was exercised only by a unit test of itself, so that test goes with it; the end-to-end test of the same behaviour stays, since `schema.coerce_slides()` and `_write_bullets()` do the clamping now. The "Known limitations" entry naming them is removed from `powerpoint.md`.

## The `promo_code` special case — removed

A `promo_code` argument used to produce a `promo_code_block` context value holding a ready-made `<div>`, in the dynamic email tool and again in the admin preview that mirrored it. Both are gone: the context is now exactly the declared arguments with `None` flattened to `""`, and nothing is synthesised.

An optional value is shown with a Mustache section over the argument itself — which is what `docs/templates.md` has shown throughout and what the example template uses. The `promo_code` *argument* is untouched; it was only ever an example name.

## Email drafts default to `en-US`

`cs-CZ` was hard-coded in two places — the tool parameter and `_create_eml_buffer()` — so every non-Czech deployment got Czech proofing on an English draft unless it patched the source.

It is `EMAIL_DEFAULT_LANGUAGE` now, defaulting to `en-US`. A Czech deployment sets the variable, which is called out against it in `docs/configuration.md` since that is the change an upgrade notices. A blank value falls back to the default rather than putting `lang=""` on a draft. The tool's `language` argument still wins per call, and `_create_eml_buffer()` takes `language=None` and resolves from config, so the default lives in one place.

## `admin/analysis.py` and `admin/preview.py`

The only modules under `admin/` with no developer page. They have a section in `dynamic-templates.md` now — what each analysis function reports, and why preview goes through the *production* substitution pipeline rather than a copy of it.

## Tests

Two new files, 31 cases; suite at 2004, `ruff` clean.

- `tests/test_upload_content_type.py` — the extension decides, a type named *inside* a name is not the type, an unknown type raises rather than guessing, and both halves of the upload layer agree on what this server generates while LibreChat still guesses where the other raises.
- `tests/test_email_language.py` — the shipped default, the environment override, a blank value, and that a per-call `language` still wins, asserted on the rendered draft's `<html lang>` and `Content-Language` rather than on config alone.

## Docs

`configuration.md`, `.env.example` and `config.py` for the new variable; `upload-backends.md` and `adding-a-tool.md` for the single MIME table and the extension rule; `powerpoint.md` for the removed limitation; `dynamic-templates.md` for the admin modules and the context rule; `docs/development/tools/email.md` for where the default now comes from.

---
_Generated by [Claude Code](https://claude.ai/code/session_01J9Jxix2j5SNrckU2u2qSit)_


dvejsada:
@claude review

Please review this PR. It closes the #116 checklist: a substring-matching `get_content_type()`, two dead PowerPoint helpers, the `promo_code_block` special case, a hard-coded `cs-CZ` email default, and two undocumented admin modules.

Worth particular attention:

- **The three behaviour changes**, each deliberate but each worth a second opinion: `.eml` becomes `message/rfc822` on the cloud backends (it was `application/octet-stream`, while LibreChat already said `message/rfc822`); `get_content_type()` now raises for a name with no known extension where a substring match used to succeed; and email drafts default to `en-US` instead of `cs-CZ`. Is any of those going to surprise someone in a way I have not accounted for?
- **The asymmetry I kept on purpose**: `get_content_type()` raises on an unknown extension, `librechat.get_mime_type()` guesses. My reasoning is in the docstrings — does it hold, or should both behave the same?
- **The `promo_code_block` removal.** No compatibility shim, by the maintainer's explicit call. Please check I removed all of it and did not leave a half-path — particularly that the admin preview and the tool body cannot now disagree about what the render context holds.
- **Dead-code removal.** `coerce_indent_level()` and `_add_image_from_url()` — confirm nothing reaches them, including through the admin UI or a dynamic tool, and that dropping the unit test did not lose coverage the end-to-end test does not have.

Same posture as #139 and #140, please: verify against the code rather than this description, and say plainly when something is pre-existing or out of scope rather than this PR's.

---

dvejsada:
Clean pass, and the checks I care about were the ones you actually ran rather than took from the description — particularly tracing `generate_named_object_name()` to establish whether the substring bug was reachable at all. I verified that end myself:

```
file_name "notes.pptx_v2" + suffix docx  ->  notes.pptx_v2.docx
   old (substring):  PPTX     <- stored in S3/GCS/Azure/MinIO as a PowerPoint
   new (extension):  DOCX
```

`sanitize_filename()` keeps dots, so a caller's dotted `file_name` reaches it. Worth adding that the failure was *ordering*-dependent, which is what made it look harmless: `Q3.xlsx.final.docx` came out DOCX before too, because the `docx` test happened to precede the `xlsx` one. Only names where the wrong type is checked first were mistyped.

On your three points, agreed on all:

- **Raise-on-unknown as "a dead branch turned into a real guard"** is a better description than mine. I had justified it by the contract; you checked there is no live path that reaches it, which is the part that makes it safe rather than merely defensible.
- **The asymmetry** — your framing (the traditional backends see only filenames this server built from a fixed suffix set; LibreChat is handed a filename it does not control) is the reason stated more precisely than the docstrings put it. I have left the code as is.
- **`promo_code_block`** — confirmed the two context builders now do only the `None` → `""` flattening, so there is no way for the tool body and the admin preview to disagree. That was the failure mode worth checking, since they had to change together.

CI is green on this head, independently of the suite numbers in the description.

Taking this as done: one pass, no findings, and the four things I flagged as most likely to be wrong were each checked against the code. Handing it to human reviewers.

---

PR #142, opened 2026-09-19
tests: cover the static email tool's draft, and fix the priority bug it found (#112)

### LINKED ISSUE #112: Add tests for the static email tool
Found while writing the developer docs for #111.

`email_tools/base_email_tool.py` (`_create_eml_buffer`, `create_eml`) has no dedicated test module. It is exercised only indirectly by `tests/test_run_blocking.py`, `tests/test_upload_unique_prefix.py` and `tests/test_librechat_integration.py`, none of which assert on the draft itself.

**Suggested `tests/test_email_creation.py`:**

- Builds a draft through `_create_eml_buffer()` and parses it with `email.message_from_bytes()`.
- Asserts headers: `To`/`Cc`/`Bcc` joined with `, `, `Subject` (RFC 2047 for non-ASCII), `Date`, `Content-Language` and `Accept-Language` from `language`, `X-Unsent: 1`, and the three priority headers for `high`/`low` and none for `normal`.
- Asserts the body is base64 `text/html; charset=utf-8` and, once decoded, contains the caller's HTML fragment unescaped, the subject HTML-escaped in `<title>`, and the sanitised language in the `lang` attribute.
- Rejects an invalid `priority`, missing `content`, missing `subject`.
- Uses the default template and a patched `find_email_template` to cover the custom-template path and the `FileNotFoundError` when neither exists.

Also worth a test: the dynamic email tool body (`dynamic_email_tools._register_single_email_template`) sets only `Subject`, recipients and `X-Unsent`, unlike the static tool. See `docs/development/tools/email.md`.

dvejsada:
Closes #112.

## The gap

`email_tools/base_email_tool.py` had no test module. It was exercised by `test_run_blocking.py`, `test_upload_unique_prefix.py` and `test_librechat_integration.py` — none of which look at the draft they produce. Everything a recipient actually receives was unasserted: the headers Outlook reads, the base64 HTML body, and which of the two caller-supplied strings gets escaped.

Two more gaps turned up behind it: the template precedence every tool relies on had no test anywhere in the suite, and the `priority` argument had a live bug.

## The bug, now fixed

The guard read `priority.lower()` and the header block further down read it again. Between them, the same category of bad input got three different answers:

| input | before | now |
|---|---|---|
| `5`, `["high"]` | `AttributeError` — raised by the guard itself, **outside** the `try`, so unlogged and unwrapped | `ValueError` |
| `None` | reached the second read → `RuntimeError` | treated as unset → `normal` |
| `0` | same route → `RuntimeError` | `ValueError` |
| `""` | quietly accepted as normal | unchanged — still normal |
| `"HIGH"` | works | unchanged |

Priority is now normalised once, at the top: unset (`None` or `""`) becomes `"normal"` — the default the MCP parameter already declares, and what `""` already did by accident — anything that is not one of the three names is a `ValueError`, and the two downstream reads compare the normalised value instead of re-deriving it.

An earlier revision of this PR left this alone as unreachable through MCP, where the parameter is typed and defaulted. That was true but not the whole story: `create_eml()` is a public entry point the package exports for direct library use, and there the argument is whatever the caller passes. Nothing else is affected — `main.py` passes the typed parameter, and the dynamic email tools don't call these entry points at all; they assemble their own `MIMEText`.

`None → "normal"` follows the convention the function already has: `language=None` resolves from config six lines below, so an explicit `None` here means "use the default", not "malformed input". Refusing one while silently defaulting the other would have been the inconsistency.

## What is covered

37 cases in `tests/test_email_creation.py`, each built through `_create_eml_buffer()` and parsed back with `email.message_from_bytes()`, so the assertions are on what a mail client reads rather than on the code's intermediates. The body is base64, so anything looking at it decodes the payload first.

| Area | Pinned |
|---|---|
| Recipients | `To`/`Cc`/`Bcc` joined with `", "`; an absent field is an absent header, not an empty one |
| Subject | A non-ASCII subject survives the round trip; an ASCII one is encoded too; the header carries the **raw** text while the `<title>` carries the escaped copy |
| Priority | The three headers move together for `high` and `low`, none of them for `normal`, case-insensitively; an unknown name, a non-string and an unset value each land where they should |
| Draft markers | `X-Unsent: 1`, a `Date`, and the language in both `Content-Language` and `Accept-Language` |
| Body | base64 UTF-8 `text/html`; the caller's fragment inserted **raw**, the subject **escaped** into `<title>` |
| Injection | A language tag cannot break out of its attribute — quotes are stripped |
| Refusals | Missing content, missing subject, empty subject |
| Templates | The shipped default; the resolved file's content reaching the render call; no template at all raising rather than producing an empty draft |
| The wrapper | `create_eml()` uploads with the right extension and filename, closes its buffer even when the upload raises, and hands a refused input back as something the caller can act on rather than an internals leak |
| The dynamic tool | Its deliberately smaller header set — no `Date`, no language, no priority |

11 further cases in `tests/test_template_resolution.py` cover **which file wins**, which is shared code rather than email's. Two independent orders decide it: the directory order (container mount before checkout, `custom_templates/` before `default_templates/`) and the candidate-name order (`custom_<x>_template` before `default_<x>_template`). Rather than pick pairs of directories to compare, the module walks the whole search list — the same filename goes into every directory and the winner is removed each round, so each boundary is exercised in turn and a fifth directory is covered the day it is added. Plus the fallback, the explicit-filename path, and the same rules serving Word and PowerPoint.

## These have teeth

Five of the six new priority cases fail on the commit before the fix, as does the wrapping test — with exactly the `'int' object has no attribute 'lower'` leak it exists to prevent. The sixth pins `priority=""`, which worked before and has to keep working; it is there so the fix cannot quietly change it, not because it was broken.

For the rest, which had no bug to regress against, each guard was checked by breaking the source. Every mutation was applied to the real source and re-run with `__pycache__` cleared and `PYTHONDONTWRITEBYTECODE=1` — see the note below on why that matters.

| Mutation to `base_email_tool.py` | Result |
|---|---|
| Join recipients with `"; "` | 1 failed |
| Drop `html.escape()` from the subject | 1 failed |
| Drop the `X-Unsent` header | 1 failed |
| Escape `content` instead of inserting it raw | 2 failed |
| Put the **escaped** subject in the `Subject` header | 1 failed |

| Mutation to `template_utils.py` | Result |
|---|---|
| Reverse the directory order (custom group after default) | 2 failed |
| Swap the **custom** pair (checkout before mount) | 2 failed |
| Swap the **default** pair only | 1 failed |
| Drop `APP_DEFAULT_DIR` from the search entirely | 2 failed |
| Reverse the `custom_`/`default_` candidates — email / docx / pptx | 1 failed each |
| Let an explicit filename fall through to the candidate list | 1 failed |

## What review found, and what it cost

Four rounds. Every finding was the same defect: **a claim that outran what a test actually exercises.**

1. **The `Subject` header.** `Header(escaped_subject, …)` passed all 29 cases here *and all 2033 in the suite*, because every subject tested was either free of HTML-special characters or inspected only through the body.
2. **The precedence claim.** `test_a_custom_template_wins` patches `find_email_template` out, so it proved only that the resolved file is the one rendered — never which file that is. The precedence itself was untested anywhere, for every tool.
3. **The fix for (2) had the same flaw, twice.** Its first draft put *differently-named* files in the two directories, so the candidate-name order decided it; the corrected draft still left `APP_DEFAULT_DIR` written by nothing, so swapping the default pair stayed green.
4. **The fix for the priority bug, too.** The `create_eml()` error-wrapping was asserted in the docs and verified by trace, with no test driving a bad priority through that entry point.

Patching each instance as it was found did not stop this — it recurred three times. Walking the list rather than picking pairs did, for the directory case.

**A measurement error worth more than any of the above.** Two of the results originally reported here were produced by stale bytecode, and one was posted as verified before it was caught. A default timestamp-based `.pyc` records the source's mtime **truncated to whole seconds** and its size, and validates on those alone; a mutation that reorders lines changes neither, so edit-run-restore inside one second silently serves the old bytecode. It fails in the flattering direction: an unapplied mutation reads as *killed*, a genuinely uncovered gap reads as *covered*. It surfaced only because a newly written test failed when it should have passed, and the runtime search order contradicted the source on disk. `docs/development/testing.md` now documents it, including that it is specific to the default `.pyc` mode.

## One thing worth a reviewer's eye

**An ASCII subject is RFC 2047-encoded too.** `Header(subject, 'utf-8')` is unconditional, so `Hello` travels as `=?utf-8?q?Hello?=`. That is legal and it is what clients get, but it looks like a bug until you check — so it is pinned by a test that says so, rather than left to surprise the next reader.

## Docs

`docs/development/tools/email.md` gets a real Tests table in place of the note saying a module was missing, the invariant that the subject reaches the draft twice in two deliberately different forms, and the priority gotcha rewritten as the fixed behaviour it now describes. Its table no longer claims to cover "the custom override", which it did not. It also loses the "Czech default locale" known limitation, which #116 removed earlier today. `docs/development/shared-modules.md` says *how* both resolution orders are covered rather than asserting that they are. `docs/development/testing.md` gains rows for reading a draft back and for monkeypatching the template directories, plus the mutation-testing note above.

Suite at 2052, `ruff` clean.


dvejsada:
@claude review

This PR is tests only, so the usual "does the change break anything" question does not apply. What I would like challenged instead:

1. **Do these assert behaviour or implementation?** A test that pins how the code happens to be written is worse than no test — it fails on a correct refactor and passes on a broken rewrite. Every draft here is built through `_create_eml_buffer()` and parsed back with `email.message_from_bytes()`, so the assertions should be on what a mail client reads. Please point at any that reach past that boundary.

2. **Is the mutation set in the PR body representative, or does it have blind spots?** Four mutations, four kills. That is evidence the tests have teeth in four specific places — it is not evidence of coverage. What could I break in `base_email_tool.py` that every one of these 29 tests would still pass on? That is the finding I actually want.

3. **`TestHeaders` pins that an ASCII-only subject is RFC 2047-encoded too**, because `Header(subject, 'utf-8')` is unconditional. I pinned the current behaviour. Is that right, or is pinning it the wrong call — i.e. should an ASCII subject go out unencoded and this be a bug report rather than a test?

4. **I deliberately left `priority=None` untested.** It raises `AttributeError` in `_create_eml_buffer()` because the validation reads `priority.lower()` twice; the MCP parameter defaults to `"normal"` so it cannot happen through the tool. A test would enshrine the wart, and fixing it is out of scope for a test-only PR. Documented as a gotcha in `docs/development/tools/email.md` instead. Disagree if you think that is the wrong trade.

Also worth a look: whether the `TestRefusals` cases cover the refusals that matter or just the ones that were easy to reach, and whether `TestTemplateResolution` actually proves the custom-override precedence rather than assuming it.

---

dvejsada:
Both findings were real. Fixed in `8bfc1b5`. I ran the mutations you could only trace statically — your reading was right on both.

**§2, the Subject header.** Confirmed empirically: `Header(escaped_subject, 'utf-8')` passed all 29 tests in this module, and all **2033** in the suite. Nothing anywhere caught it. Added `test_the_subject_header_carries_the_raw_text_not_the_escaped_one`, which fails on that mutation and passes on `master`'s code. The reason it hid is worth recording, so it is now an invariant in `email.md`: every other subject test used text `html.escape()` leaves untouched, so the two forms of the subject could collapse invisibly.

**The precedence point was the better finding.** You were right that `test_a_custom_template_wins` proves only "whatever `find_email_template()` returns gets rendered". I took your second option — a real test — because the precedence is shared by Word, Excel and PowerPoint and had no test anywhere in the suite. New `tests/test_template_resolution.py`, 9 cases.

Writing it reproduced the exact defect this series keeps producing, so I will not pretend otherwise: my first draft of the directory-precedence test put *differently-named* files in the two directories. The candidate-name order decided it and the directory order — the thing the test claimed to check — was never exercised. Reversing the directory order left it green:

| Mutation to `template_utils.py` | First draft | Now |
|---|---|---|
| Reverse directory order (custom group after default) | **survived** | killed |
| Checkout dir before container mount | killed | killed |
| `custom_email_template` / `default_email_template` reversed | killed | killed |
| Same, for `docx` | killed | killed |
| Same, for `pptx` 4:3 | killed | killed |
| Explicit filename falls through to the candidate list | killed | killed |

It now puts the *same* filename in both directories, which is the only way that axis is reachable — and is also the real case: a shipped dynamic template overridden by dropping its name into `custom_templates/`. The two orders are documented in `shared-modules.md` as independent, so the next such test is less likely to be written the same way.

**Where I did not act:**

- The `Date` `localtime` gap — agreed it is not worth a test; pinning `localtime=True` would pin implementation, not behaviour. (Small correction: the code is `localtime=True`, not `False`.)
- `TestRefusals` — agreed, "no recipients" is a product-logic question, not a test gap.
- §3 and §4 — thank you for engaging with the reasoning rather than just ratifying it. Your extension of §4 is right and is the better framing: it is *any* non-string `priority`, not just `None`, and it is one issue rather than two. I will file it that way.

`email.md`'s tests table claimed the suite covered "the custom override". It did not, and now says what it actually covers. Suite at **2043**, `ruff` clean.

---

dvejsada:
@claude review

Second round, on `8bfc1b5`. The fix for your two findings added a new 9-case module, so it deserves the same treatment rather than a pass because it was written in response to a review.

1. **`tests/test_template_resolution.py` is the new code — attack it, not the first commit.** Its first draft was already vacuous in exactly the way I then documented (differently-named files, so the directory axis was never exercised). Assume the second draft has a sibling defect I have not spotted. Specifically: the `template_dirs` fixture monkeypatches four module constants — does every path under test actually read them at call time, or does some path capture a directory at import and quietly test nothing? And does `test_the_container_mount_beats_the_checkout` really isolate the app-vs-local order, or does something else decide it?

2. **The mutation battery is my own evidence, so it is the least trustworthy part.** Six mutations, six kills — but I chose them, and I chose them *after* writing the tests. What precedence-affecting change to `template_utils.py` would all 9 cases still pass on?

3. **Does `test_the_subject_header_carries_the_raw_text_not_the_escaped_one` actually distinguish the two forms**, or does `make_header(decode_header(...))` normalise in a way that would let the escaped string through under some encoding? I verified it fails on the mutation, which proves it distinguishes *that* pair — not that the assertion is tight.

4. **Did correcting `email.md`'s tests table leave anything else overstated?** You caught one claim that outran what the tests prove. I would rather know now if there is a second.

---

dvejsada:
§2 was right, and finding it exposed that my evidence last round was partly fabricated by a tooling artifact. Both fixed in `90ed253`.

**First, a correction to what I told you.** Last round I reported a mutation table as "every one applied to the real source and re-run". Two of those cells were wrong. Python validates a `.pyc` against the source's **whole-second** mtime and its **size** — and a mutation that reorders lines changes neither. Editing, running and restoring inside one second serves stale bytecode. It fails in the direction that flatters: an unapplied mutation reads as *killed*, and a genuinely uncovered gap reads as *covered*. I only caught it because a test I had just written failed when it should have passed, and the runtime order printed `['app_custom', 'local_custom', 'local_default', 'app_default']` while the source on disk plainly read `APP_DEFAULT_DIR, LOCAL_DEFAULT_DIR`.

Re-run with `PYTHONDONTWRITEBYTECODE=1` and `__pycache__` cleared between rounds, everything is killed — but treat last round's table as unverified, not as confirmed-and-still-true. That is my error, not yours.

**§2 is real, and I confirmed it the way you couldn't.** With clean bytecode, the default-pair swap passes 10 of 11 and is caught by exactly one test — the new one. Your nine-case trace was correct: the original nine would have let it through.

**I did not take the suggested fix.** A tenth case mirroring `test_the_container_mount_beats_the_checkout` for the default group closes that boundary and leaves the next one equally unguarded — which is precisely how this gap arose, twice. So the test no longer picks pairs at all: the same filename goes into every search directory and the winner is removed each round, forcing the order to reveal itself entry by entry. A companion case asserts the fixture still knows every directory `_candidate_dirs()` returns, so a fifth one fails the suite until it is covered.

| Mutation to `template_utils.py` | Result |
|---|---|
| Reverse the directory order (custom group after default) | 2 failed |
| Swap the **custom** pair (checkout before mount) | 2 failed |
| **Swap the default pair only** — your finding | 1 failed |
| Drop `APP_DEFAULT_DIR` from the search entirely | 2 failed |
| Reverse the `custom_`/`default_` candidates (email / docx / pptx) | 1 failed each |
| Let an explicit filename fall through to the candidate list | 1 failed |

**§1 and §3** — thank you for checking these properly rather than waving them through. Your §3 reasoning is right and is the part I could not have verified by mutation alone: passing that one mutation shows the assertion distinguishes that pair, not that it is tight. That `decode_header`/`make_header` have no concept of HTML entities is the actual argument.

**§4** — `shared-modules.md` now says *how* both orders are covered rather than asserting that they are. You were right that "covers both" was the same shape of overstatement as the round-one Subject-header line: a summary that outran a specific test. Your read that `email.md:115` was narrower and accurate matches mine.

The bytecode trap is now written up in `testing.md`, including that it produced two wrong results here, since the next person mutation-testing this repo will hit it and it lies in the flattering direction.

Suite at **2045**, `ruff` clean.

---

dvejsada:
@claude review

Third round, on `90ed253`. Two rounds, two real findings, both the same shape — a test that sets up the conditions and never exercises the axis. Assume round three has one too.

1. **`test_the_whole_directory_order_is_walked_in_turn` is the new code.** It replaces pair-picking with a walk, on the theory that a walk cannot leave a boundary out. Test that theory rather than the tests: is there a precedence-affecting change to `template_utils.py` that survives all 11 cases? The walk asserts the *sequence* of resolved files — does it actually depend on the directory order, or could some other property produce the same sequence?

2. **`test_every_search_directory_takes_part_in_that_walk` compares `set(_candidate_dirs())` against the fixture's dirs.** A set, so it is order-blind by design — the walk owns order. But does it do what it claims, i.e. would it actually fail if a fifth directory were added to `_candidate_dirs()`? And is asserting against the implementation's own output circular in a way that makes it worthless?

3. **My evidence was wrong for two rounds and I did not notice.** The stale-bytecode artifact silently flattered the results. Now that I clear `__pycache__` and set `PYTHONDONTWRITEBYTECODE=1`, is there any *other* reason a mutation run in this repo could report a false kill or false survival — fixture-scoped caching, `sys.modules` state left by another test module, `monkeypatch` not unwinding, ordering dependence between test files? I would rather find the next measurement error than the next missing test.

4. **The `testing.md` write-up of the bytecode trap** — is it accurate about Python's invalidation rule? I stated it as whole-second mtime plus size. If that is wrong or incomplete, the note will mislead the next person more than no note at all.

---

dvejsada:
Clean round — no action items, and I agree with both caveats.

**The duplicate-directory mutation is inert**, as you argue. `find_file_in_template_dirs` returns the first existing match, so a repeated directory is re-checked after it has already failed and has no way to win. No observable behaviour, nothing for a test to catch. I would rather a test miss that than grow a case for it.

**Q4** — I verified the invalidation rule independently rather than take the confirmation, since I was the one who asserted it. Reading the `.pyc` header directly: `flags=0` (timestamp-based), then the source mtime stored as a 32-bit **truncated** second count and the size. Forcing a same-second, same-size edit reproduces it in isolation — source on disk reads `BBB`, the interpreter imports `AAA`. Your "truncated, not rounded" correction is the precise part and matches what I observed. One refinement I will fold into `testing.md`: naming it as the *default timestamp-based* mode, since a hash-based `.pyc` (`PYTHONPYCACHEPREFIX` aside) does not have this failure mode at all, and the note should not mislead someone who has switched.

**§3 is the most useful thing in this round** and I checked each claim rather than taking it: no `conftest.py` anywhere, neither `pytest-xdist` nor `pytest-randomly` in `requirements-dev.txt`, no `lru_cache` in `template_utils.py`. Your two "next landmine" candidates are the right ones to name — the mtime-cached PowerPoint registry, and `_reload_main_with_flag()`'s `sys.modules` popping if `template_utils` is ever added to that list. Both are the same class as the bug that actually bit me: a stale object surviving a change that looks like it took effect.

Three rounds, three real findings, and the pattern across them is worth stating plainly rather than filed away: every one was a test that set up the conditions and never exercised the axis it named. Two were mine in this PR; the third was my fix for the second. The structural answer — walking the list instead of picking pairs — only arrived after patching the individual boundary twice.

Holding here for the author's decision on merge.

---

dvejsada:
@claude review

`ba11dcd` changes what this PR is. The author asked for the latent `priority` bug to be fixed here rather than deferred to an issue, so this is no longer tests-only — it carries a behaviour change to `_create_eml_buffer()`, and that deserves the scrutiny a source change gets, not the scrutiny a test gets.

Characterising it before fixing turned up more than the earlier rounds recorded. The same category of bad input had **three** different outcomes, not one: a truthy non-string (`5`, `["high"]`) raised `AttributeError` from the guard itself — *outside* the `try`, so unlogged and unwrapped — while `None` and falsy non-strings reached a second `priority.lower()` further down and surfaced as `RuntimeError`, and `""` was quietly accepted. It is normalised once now: unset → `"normal"`, non-string or unknown name → `ValueError`, downstream comparisons against the normalised value.

What I would like attacked:

1. **Is `None` → `"normal"` the right call, or should it be a `ValueError`?** I chose it because the MCP parameter already declares `"normal"` as the default and `""` was already behaving that way, so it makes the two agree. The alternative reading is that `None` is a caller mistake and should be refused like `5` is, and that I have quietly widened what the function accepts. Argue the other side if it is stronger.

2. **`priority == ""` is the one new case that passes on the previous commit.** I kept it deliberately, to pin behaviour the fix must preserve rather than a bug it fixes. Is that a legitimate test or is it padding a table?

3. **Have I changed anything a caller depends on?** I checked that `main.py` passes the typed parameter and the dynamic email tools set no priority, so the stricter path should be unreachable in-tree. Verify that rather than take it. Note `create_eml()` wraps everything in `RuntimeError`, so a `ValueError` through *that* entry point still arrives as `RuntimeError` — same as it already did for an unknown priority string. I believe that is consistent rather than a regression; check it.

4. **Is the normalisation itself right?** `priority is None or priority == ""` then `isinstance`, then `.lower()`. Specifically: is there an input that slips through to `.lower()` and raises, or a `str` subclass that behaves unexpectedly, or an ordering problem between the three checks?

5. **Two rounds running, the defect was mine and the shape was identical.** Assume this commit has one too. The most likely place is the docs claim in `email.md`, which now describes the fixed behaviour — does it describe what the code actually does, including the `""` case and the `create_eml()` wrapping?

---

dvejsada:
Clean round, and I took the one thing you explicitly said not to block on — `f55009e`.

**§3's note was the right call to make.** Nothing drove a bad priority through `create_eml()` itself, so the wrapping claim was true by trace and unpinned by a test. You said you wouldn't block on it. I'd rather close it, because "a claim that outran what a test actually proves" is precisely the defect this PR has produced three times, and the fourth instance being small doesn't make it a different kind of thing.

The type alone distinguishes nothing — `create_eml()` wraps everything in `RuntimeError`, before and after. The **message** distinguishes it: before the fix a non-string priority reached the caller as `'int' object has no attribute 'lower'`. The test asserts on that and fails on the preceding commit with exactly the leak it exists to prevent.

**§1 is a better argument than the one I made.** I justified `None → "normal"` from the MCP parameter's declared default. You pointed at `language=None` resolving from config six lines below in the same function — so the question isn't "is `None` a mistake" but "does this function already have a convention for an explicit `None`", and it does. Refusing `priority=None` while `language=None` silently defaults would have been the inconsistency. That reasoning would have survived scrutiny better than mine; I've kept yours.

**§3's correction to my wording is right too** — I wrote that the dynamic email tools "set no priority", which understates it. They don't call `_create_eml_buffer`/`create_eml` at all; they assemble their own `MIMEText`. Different claim, and the stronger one.

**On §4**, I ran the two cases you traced but couldn't execute: `True` and `False` both reach `ValueError` (neither `is None` nor `== ""`, both fail `isinstance`), so the bool-is-an-int path you were checking for is closed. Same result as your trace.

One standing caveat on this round, not a complaint: four rounds running, you've been unable to execute anything, so every finding has been a static trace. They have all held up when I ran them — §2's short-circuit analysis and §4's type-coercion reasoning were both exactly right. But the two genuine blind spots you found in rounds one and two were *also* things a run would have shown immediately, and the round-one Subject-header finding sat behind a caveat asking someone to confirm it. Worth the repo's maintainers knowing the review is reading, not running.

Suite at **2052**, `ruff` clean. Back to the author for the merge decision.

---

Issue #160, opened 2026-09-19
Admin UI: the style-mapping editor exposes 5 of the 16 recognised keys


dvejsada:
`admin/app.py:61`:

```python
# Style-mapping keys surfaced in the UI (subset of the full recognised set).
_STYLE_KEYS = ["heading_1", "list_number", "list_bullet", "quote", "table"]
```

`docx_tools/style_map.py:60-63` recognises sixteen:

- `heading_1` … `heading_6`
- `list_number`, `list_number_2`, `list_number_3`
- `list_bullet`, `list_bullet_2`, `list_bullet_3`
- `quote`, `table`, `normal`, `code`

So a Word template that renames Heading 2, nests bullet levels under its own names, or styles fenced code blocks cannot be configured from the UI at all — and because the dropdowns show only five rows, nothing tells the admin the other eleven exist. They have to find `docs/templates.md` to learn the UI is hiding two thirds of the feature.

Fix: render all sixteen, grouped so the block stays scannable — Headings (6), Numbered lists (3), Bulleted lists (3), Other (`quote`, `table`, `normal`, `code`). The block is already inside a collapsed `Details` (`_style_mapping_block`, `app.py:540`) with a "only needed if your template renames the built-in styles" hint, so length is not a concern; the `field()` helper and a grid keep it tidy.

`_parse_style_mapping` (`app.py:315`) iterates `_STYLE_KEYS`, so it needs no change beyond the constant. `build_style_map` already handles every key and logs unknown ones, so there is no validation work either — this is genuinely just the constant and the grouping.

Issue #157, opened 2026-09-19
Admin UI: &quot;New PowerPoint template&quot; is unreachable once one template exists


dvejsada:
There is no link to `/admin/new/pptx` anywhere in the UI once at least one PowerPoint template is registered. The page works if you type the URL; nothing gets you there.

Two causes combine:

1. `_topbar` (`admin/app.py:420`) lists **All templates · New Word · New Email · Status · Log out**. There is no "New PowerPoint" entry.
2. `_template_table` (`admin/app.py:492`) renders the `Create your first …` button only in its **empty** state (`if not rows`). Once a row exists it returns a bare `Table` with no create control.

So the pptx section has a create link exactly until you use it, and then never again.

Reproduced against the live combined app with one seeded pptx template:

```
links to /admin/new/pptx: 0
links to /admin/new/docx: 2
nav: All templates | New Word | New Email | Status | Log out
```

Word and Email survive this only because the topbar covers them.

Fix: put a "New {kind}" control in the section header of every `_template_table`, in both the empty and populated states, and drop the reliance on the topbar for reachability. Add "New PowerPoint" to the nav for symmetry. A test should assert every kind in `KINDS` has a reachable create link from `/admin/` both with and without existing templates — the current shape of the bug is precisely that nobody checked the populated case.

Issue #156, opened 2026-09-19
Admin UI: no dark mode, and wide tables overflow the page on a phone


dvejsada:
Two theme gaps in `ADMIN_CSS` (`admin/app.py:67`).

**Dark mode.** The theme is already built on custom properties on `:root` — `--bg`, `--card`, `--ink`, `--muted`, `--line`, `--brand`, the status colours — so dark mode is a second block redefining those tokens under `@media (prefers-color-scheme: dark)`, not a rewrite. Everything downstream already reads the tokens. The few hard-coded literals need folding into tokens first: `#fff` on `.btn` / `input` / `.topbar nav`, `#f8fafc` on `input[disabled]`, `#eef2ff` on `.btn-secondary:hover` and `.chip`, `#f1f5f9` on `.badge-off` and `.logs td`, `#2563eb` on `.lvl-INFO`.

**Table overflow.** `.container` is `max-width:980px` with no overflow handling, and `table{width:100%}` does not stop a wide table from forcing horizontal page scroll. The status page's usage table has six columns, one of which is a free-text `last_error`; the pptx analysis table has four including a placeholder chip list. Both overflow the body on a phone. `.logs` already has `overflow:auto` and is the pattern to follow — wrap `data_table()` output in a scroll container rather than letting the page scroll sideways.

Worth doing in the same pass as the component extraction, since `data_table()` is where the wrapper belongs.

Issue #155, opened 2026-09-19
Admin UI: three duplicated fragments, two of which have already drifted


dvejsada:
Three fragments are written out twice in `admin/app.py`:

**1. The "Not found" page** — `app.py:1195-1198` and `app.py:1218-1221`, verbatim:

```python
return _page(ctx, "Not found", H1("Not found"),
             _flash(f"No managed template named '{name}'.", "err"),
             A("← Back to all templates", href=ctx.u("/")))
```

**2. The action bar** (Save & make live / Preview / Cancel) — `_edit_form` at `app.py:1035` and `_pptx_edit_form` at `app.py:790`. Already drifted: the pptx copy carries an extra explanatory paragraph and a different button label ("Preview sample deck"), which is fine as data but is currently a fork of the markup.

**3. The name field** with its `is_new` disabled/enabled ternary — `app.py:651` and `app.py:766`. Also drifted: different label ("Template name" vs "Tool name") and different hint text, again differences that belong in the per-kind descriptor, not in two copies of the same eight lines.

Extract `not_found_page(ctx, kind, name)`, `action_bar(save_label, preview_label, preview_hint=None)` and `name_field(label, hint, value, is_new)`. The drift in 2 and 3 disappears into arguments.

PR #175, opened 2026-09-19
admin: build the UI from shared components, pin the no-assets rule

### LINKED ISSUE #145: Admin UI: pin the "no external assets" property with a test that would actually catch a regression
Outcome of a review of `admin/` for external dependencies. **The UI is clean today** — every page was rendered through the real combined ASGI app and checked: no CDN stylesheet, no `<script src>`, no web font, no outward ``. The only `http://` string in the package is `analysis.py:185`, an OOXML relationship URN, not a fetch.

That is deliberate and load-bearing:

- `admin/app.py:1240` — `FastHTML(..., default_hdrs=False, htmx=False, surreal=False)` suppresses FastHTML's own pico.css / htmx / surreal CDN headers.
- `admin/app.py:67` — `ADMIN_CSS`, the whole theme inline.
- `admin/app.py:180` — `ARG_ROWS_JS`, the add/remove argument row behaviour in vanilla JS, with the blank row serialised into `window.__ARG_ROW_HTML__` at build time so no client-side templating library is needed.

The problem is the guard. `tests/test_admin_app.py:204` asserts `"cdn" not in idx.lower()` on the index page only. That would not catch `unpkg.com`, `cdnjs` under a different spelling, a Google Fonts `@import` inside a `<style>` block, or anything at all on the other five page types.

Replace it with a property test that walks every rendered admin page (index, status, new/{docx,email,pptx}, draft, edit) and asserts:

1. no `<link>` with an absolute `href`,
2. no `<script>` with a `src` attribute,
3. no absolute URL in any inline `<style>` (`@import`, `url(https://…)`),
4. no absolute `src` on any element.

Allow the `<link rel="canonical">` FastHTML emits, which is same-origin by construction. Parsing with `beautifulsoup4` is fine — it is already a runtime dependency.

dvejsada:
Closes #145 and #146 with all five of its sub-issues (#152, #153, #154, #155, #156).

## What was wrong

`admin/app.py` held the theme, the view fragments, the form parsing, every route and the app factory in 1,377 lines. A fragment layer existed, but the primitives below it did not, so the same markup was written out by hand:

| Pattern | Occurrences |
|---|---|
| `Div(..., cls="card")` | 19 |
| `Div(Label(…), <control>, P(cls="muted"), cls="field")` | 23 |
| `Table(Thead(Tr(Th…)), Tbody(*rows))` | 3 |
| checkbox-inside-label | 3 |
| `if kind == KIND_PPTX` | 15 |

Three fragments were duplicated outright — the "Not found" page, the action bar, the name field — and two had already drifted apart.

## What this does

```
admin/components.py   markup primitives + the inlined theme
admin/kinds.py        one KindDescriptor per template kind
admin/forms.py        reading a submitted form back into a spec
admin/views/          shell, templates, status, login
```

`app.py` is now routes and `AdminContext` only, **1,377 → 422 lines**. `components`, `kinds`, `forms`, `store` and `analysis` do not import `app.py`; views take the `AdminContext` as a parameter, and `store`/`analysis`/`forms` stay FastHTML-free so their rules are unit-testable without rendering.

## Three fixes the extraction makes possible

**Not one label in the UI was associated with its control** (#153). Labels were bare siblings, controls had no ids — a screen reader announced an unlabelled box, and clicking a label focused nothing. `field()` mints an id from the control's `name` and points the label at it, fixing all 23 at once. `<html>` gains `lang="en"`. Verified on the rendered pages: 24 label/control pairs, no duplicate ids, no orphan `for` targets.

**Per-kind metadata was split across two modules** (#154), contradicting `store.kind_meta`'s own docstring that adding a kind means editing one table. `KindDescriptor` re-exposes the storage fields, so a view needs one lookup, and the pptx branches become attribute reads. Where behaviour genuinely differs it now branches on `has_args` — the real distinction — rather than the kind name.

**The theme is now token-based** (#156): colours are custom properties on `:root`, redefined under `prefers-color-scheme: dark`, so the UI follows system appearance. `data_table()` wraps its table in a scroll container — the status page's six columns previously scrolled the whole document sideways on a phone.

## The test (#145)

The old guard was `assert "cdn" not in idx.lower()` on the index page alone. That passes happily for `unpkg.com`, a Google Fonts `@import`, or anything at all on the other five page types.

`tests/test_admin_assets.py` walks all ten page types (including the signed-out pages and the two that only exist after an upload) and rejects: off-origin `<link>`, any `<script src>`, `@import`/`url()` reaching off-origin in inline CSS, and off-origin `src`/`srcset`/`poster`/`data` on any element. It also asserts the theme and script **are** inline, so deleting them is not a way to pass.

Verified by mutation — each of these fails it, and passes again on revert:

| Mutation | Result |
|---|---|
| CDN stylesheet `<link>` | 1 failed |
| `<script src="unpkg.com/htmx.org">` | 2 failed |
| Google Fonts `@import` in the inline CSS | 1 failed |
| Off-origin `` | 1 failed |
| Dark-mode block deleted | 2 failed |
| `lang` removed from `<html>` | 1 failed |

## Verification

- `ruff check .` clean; `pytest -m "not network"` → **2058 passed**.
- Behaviour checked by rendering every page from this branch and from `master` and diffing the HTML. The only differences are the intended ones: `lang="en"`, the `for`/`id` pairs, and the `table-wrap` wrappers. That diff also caught three copy regressions I'd introduced (a dropped "templates", a duplicated label, a dropped "New") before they reached the commit.

**One deliberate behaviour change.** The "Replace document" file picker on a pptx template accepted only `.pptx`, while the create page accepted `.pptx` and `.potx` and the reupload route validated neither. Both now use the kind's full extension list. Shared components are what surfaced the inconsistency; say the word if you'd rather keep the old asymmetry.

## Deliberately left alone

PowerPoint's missing "New template" nav entry (#157) and the style-mapping key subset (#160) are untouched, with the deciding constants named (`kinds.NAV_KINDS`, `kinds.STYLE_KEYS`) and noted in the docs so those fixes stay reviewable on their own.

## Docs

New `docs/development/admin-ui.md` — module map, the three invariants, how to add a kind, known limitations — linked from `docs/README.md`. `AGENTS.md` gets the corrected layout map and an Admin UI rules block. `docs/admin-ui.md` notes dark mode and the offline guarantee.

---


dvejsada:
@claude review

This is a large mechanical refactor, so the things most worth an independent check are:

1. **Behaviour equivalence.** `admin/app.py` went 1,377 → 422 lines with the views, primitives, per-kind table and form parsing moved out. I verified this by rendering every page on this branch and on `master` and diffing the HTML — the only differences are `lang="en"`, the `for`/`id` pairs, and the `table-wrap` wrappers. Please look for anything that diff would not catch: changed POST handling, a route that now returns a different status, lost validation in `draft`/`reupload`/`save`, or a spec key that `admin/forms.py` writes differently from the original `_build_spec`.

2. **The one intentional behaviour change.** The pptx "Replace document" picker previously accepted only `.pptx` while the create page accepted `.pptx` and `.potx`, and the reupload route validated neither. Both now use the kind's full extension list. Worth a second opinion on whether that's the right resolution.

3. **`components.field()`.** It mutates `control.attrs["id"]` in place to wire up the label. Please check that for aliasing hazards — a control object reused across calls, or a caller that passes its own `id`.

4. **The new test's blind spots.** `tests/test_admin_assets.py` asserts no off-origin `<link>`, `<script src>`, CSS `@import`/`url()`, or element `src`. I mutation-tested six violations and all six fail it. What it would still miss is more interesting than what it catches.

---

dvejsada:
Good review — the main finding was real, and chasing one of the test notes turned up something worse. All four addressed in 13504f3.

**1. The reupload-reject page.** Confirmed against master: its error branch rendered `_page → _flash → _edit_form → _replace_card`, no `_analysis_report`, while computing `cur` purely for the arg-row prefill. Reverted to match rather than kept.

The reasoning, since "plausible improvement" was a fair reading: on the reject path the file on disk is still the *old* one, so it is the right thing to prefill argument rows from and the wrong thing to put under "What we found in the document" — a heading that means "the file you just uploaded" on every other path it appears. Same card, two meanings, nothing on screen distinguishing them. `edit_page()` now takes `report`, and the reject path asks for prefill without report.

You also spotted only half of it. The same `edit_page()` branch renders a "source file is missing" warning when `analysis is None`, which master's reject path likewise never showed. `report=False` suppresses both.

Tests added for the whole path, which had none — empty, unopenable and oversized uploads, that the installed file survives a rejection, and that the card shows on the edit page and on a successful re-scan but not on a rejection.

**2. Login CSRF field.** Fixed. It came from routing login through `post_form()`, whose token defaults to `""`. Plain `Form` now, with the reason in a comment: there is no session yet to hold a token, so an empty field implies a protection the route does not have. Pinned by a test.

**3. `field()` aliasing.** Agreed on both counts — nothing trips it today, and it is worth naming. The docstring now states that the `id` is set in place and what reusing one control instance across two calls would do. No behaviour change.

**4. The two test blind spots — and a third underneath them.** Both correct and both fixed: `style=` attributes are now scanned, and `srcset` is split into candidates before matching.

Fixing the first exposed the real problem. **My style-attribute check was vacuous.** The sweep only ever built docx templates, so no page in it rendered a `swatch()` — I poisoned `swatch()` with `style="background:url(https://evil.example/x.png)"` and all seven tests passed. The assertion was fine; it had nothing to look at.

So the sweep now covers the PowerPoint draft and edit pages too (12 pages, up from 10), and there is a test asserting the sweep actually reaches a swatch — so this check cannot go vacuous the same way again. Both mutations fail it now.

That is the second time on this PR that verification-of-the-verification caught something the verification missed, which is a fair argument that my original "six mutations, all caught" claim was worth less than it sounded.

**On my own PR description:** it claimed the only rendered diffs against master were `lang`, `for`/`id` and `table-wrap`. That was wrong — the reject page and the login form were both real deltas my page-by-page diff never covered, because it only walked pages reachable by GET plus the two happy-path POSTs. With both reverted the claim is true again, but the method had a hole and the description overstated it.

`ruff` clean, **2066 passed**.

---

Issue #159, opened 2026-09-19
Admin UI: the auth gate exempts any path ending .css or .ico


dvejsada:
`admin/auth.py`, in `make_before._before`:

```python
if path in allowed or path.endswith(".ico") or path.endswith(".css"):
    return None
```

Any request whose path ends `.css` or `.ico` bypasses the session check entirely.

**Nothing leaks today.** The exemption was presumably written for static assets, but the admin UI serves none — the theme is inlined at `app.py:67` and the JS at `app.py:180`, which is the whole point of the no-external-assets design. I checked the route table and no current route can match a path with those suffixes, so every exempted request falls through to a 404.

That is the problem: the exemption buys nothing and waits. `/{kind}/{name}/edit` takes a free-form `name`; the day a route is added whose path can end in `.css` — a download endpoint, an export, a static mount for a logo — it is unauthenticated, and nothing in the code or tests says so.

Fix: drop both suffix checks. If static files are ever served from the admin mount, exempt that mount's prefix explicitly rather than a suffix pattern. Add a test that a request to an arbitrary `/admin/*.css` path redirects to login rather than being served.

Issue #158, opened 2026-09-19
Admin UI: deleting a template silently orphans its source file, with no way to see or clear it


dvejsada:
`admin/app.py:1351` hard-codes the asset to be kept:

```python
ctx.store.delete_spec(kind, name, delete_asset=False)
```

`FileTemplateStore.delete_spec` supports `delete_asset=True`; the UI never offers it. The confirmation text does say the source file is kept in `custom_templates/` — but nothing in the UI ever lists that directory, so a kept file is a file you cannot subsequently see, inspect or remove without shelling into the volume.

Delete a few templates over a few months and `custom_templates/` holds an invisible pile of stale `.docx` / `.pptx` / `.html` files. Worse, the names collide with intent: re-creating a template with the same name silently reuses whichever old asset is still sitting there, because `save_spec` derives the filename from the template name and only raises if the asset is *missing* (`store.py`, `save_spec`).

Fix: replace the hard-coded `False` with a choice on the confirmation — "also delete the source file" — defaulting to keeping it, which preserves today's behaviour for anyone who wants it. Make the consequence explicit in the confirm text either way.

The complementary half, a view that lists assets in `custom_templates/` no spec points at, is filed separately under the lifecycle issue — this one is just about not creating orphans silently in the first place.

PR #176, opened 2026-09-19
admin: fix the three defects from the component review

### LINKED ISSUE #147: Admin UI: defects found in the component review
Three defects found while reviewing `admin/` for shared components. They are independent; separate PRs are fine.

One is a dead end in the navigation that makes a whole template kind unreachable, one silently accumulates orphaned files on disk, and one is a latent auth footgun that leaks nothing today only because no route happens to match.

A fourth finding from the same review — the style-mapping editor exposing 5 of the 16 recognised keys — is filed under the capability-gap issue instead, since it is a missing capability rather than a malfunction.

Sub-issues below.

### LINKED ISSUE #166: Admin UI: a view of the files in custom_templates/ that no template points at
Nothing in the UI ever lists `custom_templates/`, so the directory is write-only from the admin's point of view: files go in through uploads and are never shown again.

That leaves orphans invisible. Deleting a template keeps its source file by design (`app.py:1351`), a rename or a replace can strand the old one, and files copied onto the volume by hand are equally unlisted. Nothing reports them and nothing cleans them up.

There is a second-order hazard: `save_spec` derives the asset filename from the template name and only raises when the asset is **missing**, never when one already exists. So creating a template whose name matches a previously deleted one silently adopts whatever stale file is still sitting there — with no indication in the UI that a file was reused rather than uploaded.

Add a maintenance view listing every file in `custom_templates/` with its size and mtime, marked as: referenced by a managed spec, referenced by a master-YAML template, or unreferenced. Offer deletion for the unreferenced ones only, one at a time, with the filename in the confirmation.

Read `template_registry.gather_specs` for the master-YAML side so hand-written templates are not mistaken for orphans — that is the failure that would make this feature actively dangerous.

Complements the delete-with-asset choice filed under the bugs issue: that one stops *new* orphans, this one deals with the ones already there.

dvejsada:
Closes #147 and its three sub-issues (#157, #158, #159). Follows #175, which is why these were held back — two of the three are one-line changes to constants that PR deliberately left named.

## #157 — PowerPoint's create page was unreachable

The top bar listed New Word and New Email but not New PowerPoint, and `template_table()` rendered a create link only in its **empty** state. So registering one PowerPoint template removed the only route to the page that made it. Word and Email never showed the bug because the top bar covered them.

Fixed on both sides, deliberately redundantly: `NAV_KINDS` is every kind, and the table renders a create link in both states.

The test pins each half separately. That is not belt-and-braces for its own sake — either fix alone satisfies "the page is reachable", so a single assertion would let the other rot silently. Confirmed: with only the table link restored, dropping PowerPoint from the nav still passed until I added the nav assertion.

## #158 — deleting a template always kept its source file

`delete_asset` was hard-coded `False` with no way to choose, and nothing in the UI lists `custom_templates/` — so a kept file is one you cannot afterwards see or remove. There is a sharper edge behind it: re-creating a template under a previously used name silently adopts whatever stale asset is still sitting there, because `save_spec` only objects when the asset is *missing*.

A JavaScript `confirm()` can't hold a checkbox, so the confirmation became a page — same shape as the existing sign-out confirm — that names what is about to go and offers the file as an opt-in. Keeping it stays the default, so today's behaviour is still one click away.

It also **refuses to delete an asset another managed template points at**. Assets share one flat directory, so two specs can name the same file (usually because one was hand-written). The page withholds the option, and the POST ignores the flag even if it arrives anyway — the page hiding it is not where that decision should be trusted from.

This does not close #166; nothing yet lists the orphans already on disk.

## #159 — the auth gate exempted any path ending `.css` or `.ico`

Removed. Nothing leaked, and investigating it showed the exemption was **more inert than the issue claimed**: the admin UI serves no static files, so no route matches such a path and the request 404s before `before` is consulted at all.

That also changes how it can be tested. My first attempt asserted a 303 from `GET /admin/theme.css` and failed with a 404 — an HTTP-level test can't reach the code this changes. It's now a direct test of `make_before()`: login is public, everything else redirects, an authenticated session passes and gets a CSRF token.

## Verification

`ruff` clean, **2073 passed**. Each fix mutation-tested against its own unfixed code — all five fail the suite:

| Mutation | Result |
|---|---|
| Restore the `.css`/`.ico` exemption | 1 failed |
| Drop pptx from `NAV_KINDS` | 1 failed |
| Create link only in the empty state | 1 failed |
| Re-hard-code `delete_asset=False` | 1 failed |
| Trust the posted flag for a shared asset | 1 failed |

## Docs

`docs/admin-ui.md` describes the new delete step. `docs/development/admin-ui.md` drops the two limitations that are gone, narrows the third to what remains (#166), and records why the create link is duplicated in two places — that redundancy looks removable and isn't.

---


dvejsada:
@claude review

Three independent defect fixes. The places I'd most want a second pair of eyes:

1. **The delete flow is the only one with real teardown semantics.** `delete()` is now GET-confirms / POST-deletes on one route. Please check the POST path for anything the GET guards but the POST doesn't — particularly that `shared_with` is recomputed on the POST rather than trusted from the form, that a missing spec is handled on both methods, and that CSRF still gates the destructive half.

2. **`other_specs_using_asset()`** compares bare filenames across all kinds to decide whether an asset is shared. Is that the right comparison? Consider a spec whose path key is absent or empty, a master-YAML template pointing at the same file (which this does *not* check — only managed specs), and whether a file in `default_templates/` rather than `custom_templates/` could be wrongly treated as deletable.

3. **`#159`'s test is a unit test of `make_before()`, not an HTTP test** — because unmatched paths 404 before the gate runs, so nothing at the HTTP level can exercise it. Please sanity-check that reasoning; if there *is* a reachable route whose path can end `.css` or `.ico`, the exemption was live and the test should be at that level instead.

4. Anything in the delete confirmation page that could mislead about what is about to happen — that page is the last thing between a user and a destroyed template.

---

dvejsada:
The master-YAML gap was real and is fixed in 09d25b4 — and it had a second half worth naming.

**The gap.** Confirmed: `other_specs_using_asset()` only walked `store.list_specs()`, which reads the managed `*.d` directories. The master YAML is a separate source the UI already renders as read-only rows, and it points at the same flat `custom_templates/`. So the page offered the option and the POST honoured it. Fixed rather than documented — the master YAML is precisely where a hand-written duplicate comes from, which is the case this check exists for, so a limitation note would have described the check failing at its one job.

**The half you didn't raise.** A master entry sharing the *name* being deleted also counts. The managed spec was overriding it, so deleting the override doesn't retire the name — it revives the hand-written definition, still pointing at that asset. Deleting the file there breaks the template that just came back. Both cases are now tested, and the mutation that skips only same-name master entries fails.

`KindDescriptor` gained `master_file`, derived from the spec subdirectory, so the master YAML's location stays in the kind table rather than being rebuilt at the call site.

**On your other three points** — agreed on all, no changes needed:

- **Delete flow.** Your reading matches mine: `is_kind` / `get_spec` / `shared_with` all resolve before the method branch, `shared_with` is re-derived server-side on the POST, CSRF gates only the mutating half.
- **`other_specs_using_asset` non-issues.** Right on both: the empty-path-key short-circuit can't false-positive, and `delete_spec` unlinks only via `asset_path()`, which resolves inside `custom_dir` — `default_templates/` is unreachable regardless of the verdict.
- **The `.css` test placement.** Your check agrees with what I found the hard way: my first attempt *was* an HTTP test, and it failed with a 404 rather than the redirect it asserted. No route's path can end `.css` or `.ico`, so the gate is never consulted for one, and the unit test is the only level that reaches the changed line.

I took the CSRF coverage note even though it was phrased as optional — it's one assertion on the only route that destroys anything, and `test_post_without_csrf_is_rejected` covered `/save` alone. The mutation that drops `_csrf_guard` from `delete` now fails.

`ruff` clean, **2076 passed**. Three new mutations verified: not consulting the master YAML, skipping a same-name master entry, and removing the CSRF guard each fail the suite.

---

Issue #163, opened 2026-09-19
Admin UI: show the YAML spec the UI writes


dvejsada:
The UI writes one YAML file per template into `config/<kind>_templates.d/` and never shows it. `FileTemplateStore._dump_spec` even stamps a header saying the file is UI-managed and that edits are merged on top of the master YAML — a message aimed at someone reading the file on disk, which the UI gives no way to do.

That matters more here than it would in a typical admin tool, because the YAML *is* the product's documented interface. `docs/templates.md` teaches templates as YAML; `config/*.yaml` are hand-written and heavily commented. An admin who builds a template by clicking has no way to see the artefact in the vocabulary the rest of the documentation uses, which makes the UI and the docs feel like two unrelated products.

Concretely it blocks: reviewing what a form actually produced before saving, pasting a working spec into the master YAML or another environment, filing a useful bug report, and learning the YAML format from the UI rather than instead of it.

Add a read-only YAML view on the edit page — a collapsed `Details` block rendering `FileTemplateStore._dump_spec(spec)` in a `<pre>`. Read-only on purpose: an editable YAML box means round-tripping arbitrary input through `save_spec` and reconciling it with the form, which is a much larger change and not what this is for.

Pairs naturally with an export route (the same bytes as a `.yaml` download) if the source-file download lands at the same time.

Issue #162, opened 2026-09-19
Admin UI: no way to download the source file that is actually installed


dvejsada:
The edit page offers "Replace document" (`_replace_card`, `admin/app.py:874`) but no way to fetch what is there. So an admin can overwrite a template's source file but cannot:

- check which version is actually deployed before replacing it,
- copy a working template to a second environment,
- recover the file after losing the local original,
- diff what the UI is rendering against what they think they uploaded.

The only route to the bytes is shelling into the volume — which is the friction `docs/admin-ui.md` says the UI removes.

Everything needed is already in place: `FileTemplateStore.read_asset(kind, filename)` and `asset_exists`, with `validate_asset_filename` enforcing a bare filename so a download route cannot be walked out of `custom_templates/`. Add a GET route returning the bytes with the right media type and a `Content-Disposition` attachment, and a Download button next to Replace.

Two things to get right:

- The route must sit **behind** the auth gate. Note that the gate currently exempts paths ending `.css`/`.ico` (filed as a separate bug) — a download route serving admin-uploaded filenames is exactly the kind of route that turns that latent exemption into a real one, so land the auth fix first or together with this.
- Resolve the filename through `store.asset_path()` rather than joining it by hand, so the traversal guard applies.

Issue #161, opened 2026-09-19
Admin UI: the global style_mapping is read-only


dvejsada:
`docs/templates.md` documents two levels of style mapping: a top-level `style_mapping:` in `config/docx_templates.yaml` that applies to **every** conversion, and a per-template one that overrides it.

The UI can edit the per-template mapping. The global one it only reads — `AdminContext.global_style_mapping` (`admin/app.py:202`) re-reads the master YAML on every access and feeds it to `build_style_map` for previews, but there is no view and no form. So the setting with the widest blast radius, the one that rebrands every Word document the server produces, is the one that still requires hand-editing YAML on the volume.

It is also invisible: an admin looking at a per-template style dropdown has no way to see what the global mapping already sets, so "(use built-in)" is a lie whenever a global override is in force — it means "use the global mapping, whatever that is".

Two parts, and the second is worth doing even alone:

1. **Show it.** On the per-template style block, label each "(use built-in)" option with what it actually resolves to — `(inherit: Brand Title)` when the global mapping sets `heading_1`. That removes the lie at no storage cost.
2. **Edit it.** A settings page for the global mapping, writing to the same `*.d` merge layer the per-template specs use rather than touching the hand-written master YAML — consistent with the store's stated rule that master files are never modified. Needs a home for global config in the merge layer, which `gather_specs` already returns as the second element of its tuple.

PR #177, opened 2026-09-19
admin: close the YAML/UI capability gap

### LINKED ISSUE #148: Admin UI: close the gap between what template YAML can express and what the UI can edit
`docs/admin-ui.md` sells the UI as "prefer clicking over editing YAML?", but several things the master YAML can express have no control in the UI at all, so an admin who needs one of them is back in a text editor — and, worse, has no way to see from the UI that the setting exists.

The four gaps that come up in practice: the style-mapping editor covers under a third of the recognised keys, the global (all-documents) style mapping is read-only, there is no way to fetch the source file that is actually installed, and there is no way to see the YAML the UI writes.

None of these need new storage — they are all views onto data the store and `template_registry` already hold.

Sub-issues below.

dvejsada:
Closes #148, #160, #162 and #163. **#161 stays open** — half of it is done and the other half doesn't belong here; reasoning below.

## #160 — the style editor offered 5 of the 16 keys the renderer acts on

A Word template that renames Heading 2, nests bullet levels under its own names, or styles fenced code simply could not be configured, and because the panel showed five rows nothing hinted the other eleven existed.

All sixteen now, grouped into **Headings / Numbered lists / Bulleted lists / Other blocks**.

`tests/test_admin_style_keys.py` compares `kinds.STYLE_KEYS` against `style_map`'s own `_HEADING_KEYS` / `_LIST_NUMBER_KEYS` / `_LIST_BULLET_KEYS` / `_SCALAR_KEYS` in **both** directions — offering fewer hides the feature, offering more promises a setting that silently no-ops. Adding a key to the renderer now fails the suite until the UI catches up.

## #161 — half done, and why only half

Each dropdown's default option read `(use built-in)`. That is untrue wherever the global `style_mapping` sets the key: there it means "use the global mapping", and nothing on screen said what that was. It now reads `(inherit from global: Brand Title)` or `(use built-in: Heading 1)`, with the built-in names read off `DEFAULT_STYLE_MAP` rather than restated, so the labels can't drift from what a document actually gets.

**Editing** the global mapping is not here. It needs:

1. a home for global config in the `*.d` merge layer — `gather_specs()` reads top-level keys from the master YAML only, and the store's rule is that master files are never rewritten; and
2. `style_map.load_global_style_map()` to stop caching per-process, or an edit won't take effect without a restart.

Both are core-module changes that also affect the **static** Word tool's styling — a different risk profile from three additive UI changes, so I've left #161 open with that written down rather than bundling it. Happy to take it next if you'd rather.

## #162 — no way to fetch the installed source file

You could replace it but never read it, so you couldn't check what was deployed, copy a template to another server, or recover a lost original. Adds a download route and a button beside the upload.

It resolves through `store.asset_path()`, whose `validate_asset_filename` refuses anything but a bare name. `save_spec` validates on write, but a `*.d` spec is plain YAML someone can hand-write — so there's a test that plants `docx_path: ../secret.txt` and asserts the file is not served. Joining the path by hand instead fails it.

This is also the first route to serve an admin-uploaded file, which is why #159's gate fix landing first mattered; a test asserts the route redirects to login when signed out.

## #163 — the YAML was never shown

`FileTemplateStore.dump_spec` (made public for this) rendered read-only on the edit page. The YAML *is* the documented interface — `config/*.yaml` is hand-written and `docs/templates.md` teaches templates in it — so a template built by clicking should be readable in that vocabulary. Read-only on purpose: round-tripping arbitrary YAML back through `save_spec` is a much larger change.

## Verification

`ruff` clean, **2089 passed**. Mutation-tested:

| Mutation | Result |
|---|---|
| Back to the 5-key subset | 3 failed |
| Label ignores the global mapping | 1 failed |
| YAML block removed | 1 failed |
| Download link removed | 1 failed |
| Download joins the path by hand | 1 failed |

One mutation I tried **passed**, and it was my mistake rather than a gap: I removed the "asset missing" guard in the download route, but `read_asset` raising already covers that case, so behaviour was unchanged. The property actually worth pinning there is traversal, which is what the `../secret.txt` test does.

## Docs

`docs/admin-ui.md` covers download, the YAML view and the full style panel. `docs/development/admin-ui.md` drops #160, restates what remains of #161 with what it would take, and records the drift guard. `AGENTS.md` gains the style-key rule.

---


dvejsada:
@claude review

Four changes; the ones I'd most want checked:

1. **The download route** (`admin/app.py`) is the first thing in this UI to serve an admin-uploaded file. Please check it end to end for path handling — a spec's `docx_path`/`pptx_path` is plain YAML a person can hand-write, so `save_spec`'s validation is not in the path. Is resolving through `store.asset_path()` sufficient, and is there any input (absolute path, symlink in `custom_templates/`, odd extension, a name that is a directory) that gets past it? The `Content-Disposition` filename is also interpolated from the spec.

2. **`kinds.STYLE_KEYS` vs `style_map`.** `tests/test_admin_style_keys.py` reaches into `style_map`'s private `_HEADING_KEYS`/`_SCALAR_KEYS` tables to compare. Is that the right coupling, or does it just move the drift risk? And does `builtin_style_names()` reconstruct the key→default mapping correctly for every key, particularly the tuple-indexed list levels and `code` (whose default is `None`)?

3. **`#161` is deliberately half-done.** I've left editing the global mapping out because it needs a merge-layer home for global config *and* un-caching `load_global_style_map()`, both of which affect the static Word tool. Please sanity-check that reasoning — if there's a contained way to do it that I've missed, say so.

4. **`dump_spec` went from private to public** and is now rendered into a page. Anything in a spec that could make that output misleading or unsafe to display?

---

dvejsada:
Good catch on the header, and it was worse than reported. Fixed in 04b47f8.

**1. The `Content-Disposition` filename.** You flagged the quote breakout. Checking it turned up that `validate_asset_filename` also accepts CR/LF:

```
ACCEPTED: 'evil".docx'
ACCEPTED: 'a\r\nX-Injected: 1.docx'
```

So the same path was header *injection*, not just a broken filename — which moves it past "cheap defence in depth".

Fixed at both ends rather than one:

- `validate_asset_filename()` rejects quotes, backslashes and control characters. Deliberately **not** spaces or non-ASCII: `Brand Deck.pptx` and `naïve.docx` are ordinary names, and outlawing them would break existing deployments for no security gain. That's why I didn't take the "reject non-alphanumeric" suggestion as written.
- `kinds.content_disposition()` builds the header per RFC 6266. Worth noting the old f-string wasn't only injectable — it couldn't legally encode a non-ASCII name at all. It now emits an ASCII-folded `filename=` plus `filename*=UTF-8''…`, with a test that downloads a template named `smělouva.docx` and asserts the header is ASCII-encodable.

Agreed on your other two download points: `read_asset`'s `is_file()` check makes a directory-name collision fail closed, and `attachment` over `inline` is deliberate for exactly the `.html` reason you give.

**2. `RECOGNISED_KEYS`.** Taken, though you didn't block on it — `docx_tools/style_map.py` now exports it and the test compares one name instead of unioning four private tables. The mutation that adds `caption` to `_SCALAR_KEYS` fails the suite, so the guard demonstrably still fires through the new indirection.

**3. #161.** Thanks for checking both blockers independently rather than taking the PR description's word — particularly the point that the caching is keyed at module import, so it can't be worked around from the admin layer. Nothing changed.

**4. The YAML block's escaping.** You flagged that you couldn't execute code and were inferring from the codebase's pattern. I can, so I checked rather than leaving it inferred:

```
>>> to_xml(Pre(Code('desc: <script>alert(1)</script>')))
<pre><code>desc: &lt;script&gt;alert(1)&lt;/script&gt;</code></pre>
```

It escapes. There's now an end-to-end test that saves a description containing `<script>alert(1)</script>` and asserts the edit page renders it escaped — the inference was right, but this is the kind of claim worth a test rather than a reasoned argument.

`ruff` clean, **2099 passed**. Three new mutations verified: dropping the character guard, reverting to the raw f-string header, and adding a renderer key the UI doesn't offer each fail the suite.

---

Issue #170, opened 2026-09-19
Admin UI: Excel has no presence at all; surface the named styles a workbook template defines


dvejsada:
`admin/` has no Excel anything: `KINDS` is `(docx, email, pptx)`, `analysis.py` dispatches three kinds, the index page has three sections.

Excel does have a template — `custom_xlsx_template.xlsx` — and it carries a genuinely awkward piece of configuration. Per `docs/templates.md`, any named cell style in that workbook becomes usable as `style:<Name>` in a `styles:` directive. So the set of valid style names is defined by a binary file on a volume, and there is no way to ask the running server what it is. Getting one wrong is a silent no-op or a fallback, and the only way to learn the correct spelling is to open the workbook in Excel and go looking through the styles gallery.

That is exactly the question an admin UI should answer, and the analysis machinery for it is trivial — `openpyxl` is already a dependency and `Workbook.named_styles` gives the list directly.

Scope here: an Excel card on the base-templates page that, alongside upload/download/revert, lists every named style the installed workbook defines, so an admin can read off the names that `style:` directives will accept. Flag the ones that collide with Excel's own built-ins if that turns out to matter in practice.

This does not make Excel a dynamic-template kind — there are no xlsx placeholder templates and no `xlsx_templates.yaml`. It is one more card on the static page, plus a fourth entry in whatever per-kind descriptor table the component work lands on.

Depends on the base-templates page existing.

dvejsada:
Shipped in #178 (merged as `ab16419`), as part of the base-templates page this depended on.

The Excel card lists every named style the installed workbook defines, so the values a `styles:` directive will accept as `style:<Name>` can be read off the page instead of by opening the workbook in Excel. `analyze_xlsx()` puts them in `Analysis.styles_present` — the same field that means "the style names this document offers" for the Word template — and a test asserts the page and `xlsx_tools.styles.load_template_styles()` agree, so the UI can't drift from what the tool actually accepts.

Two notes on what did and didn't happen:

- **`Normal` is excluded**, matching `load_template_styles()`, which doesn't offer openpyxl's built-in default as a referenceable name.
- **Built-in collisions are not flagged.** The description made that conditional ("if that turns out to matter in practice"), and nothing so far suggests it does — worth its own issue if a real workbook turns up where it bites.

[...]

PR #178, opened 2026-09-19
admin: a page for the static base templates

### LINKED ISSUE #169: Admin UI: a page for the static base templates (docx, email, pptx 4:3 + 16:9, xlsx)
The five static templates in `docs/templates.md` — `custom_docx_template.docx`, `custom_email_template.html`, `custom_pptx_template_4_3.pptx`, `custom_pptx_template_16_9.pptx`, `custom_xlsx_template.xlsx` — decide how **every** generated document looks, and the admin UI manages none of them. Changing one means shelling into the volume and copying a file.

This is probably the single biggest usability gap in the feature. The dynamic-template flow is the advanced case; dropping in a branded Word template is the first thing most people want, and it is the one thing the UI cannot do.

A "Base templates" page, one card per slot:

- **State** — whether a custom file is installed or the bundled default is in use. `template_utils` already resolves custom-then-default and logs which it picked; the same resolution drives the display.
- **Upload / replace**, with the same size and format validation the dynamic flow uses, and the same "analyse and report" treatment — the docx style check from `analysis.analyze_docx` (which styles the renderer needs are missing) and the pptx layout/role/theme report from `analyze_pptx` are exactly as useful here as they are for a named template, arguably more so.
- **Download** the installed file.
- **Revert to default** — delete the custom file so resolution falls back to `default_templates/`. Needs a clear confirmation; this is the destructive one.
- **Preview** — render a short sample through the template, reusing the docx and pptx preview paths already in `admin/preview.py`.

Two things to work out in design review: these files are referenced by fixed filename rather than by a spec, so `TemplateStore` needs a path for assets with no spec behind them; and any caching in `template_utils` has to be invalidated on upload the way `pptx_tools.templates.clear_cache()` is today, or a replaced base template will not take effect until restart — which would quietly break the "no restart" promise the UI is built on.

Excel's own facet (listing the named cell styles the uploaded workbook defines) is filed separately.

dvejsada:
Closes #169 and #170.

The five files that style **every** generated document — the Word template, the email wrapper, both PowerPoint designs and the Excel named-styles workbook — could only be changed by shelling into the volume and copying a file. They're the most commonly customised files in the product and the UI managed none of them.

## The page

One card per slot: which file is in use (yours or the bundled default), **download** it, **replace** it, **revert**.

A replacement takes effect on the next document with no restart — and unlike the named-PowerPoint registry there was no cache to invalidate: `template_utils` resolves per call and `xlsx_tools.styles.load_template_styles()` reloads per call. I checked that before building on it rather than assuming.

Uploads are validated before anything is installed — size and emptiness (in `_read_upload`), then extension against the slot, then openability via `admin.analysis`. A rejected upload leaves the installed file untouched, which is tested.

Each slot reports what its file offers in the terms that matter for its kind: the Word template's styles **and which ones the renderer needs but cannot find**, the wrapper's Mustache variables, a deck's layouts/roles/theme, and the workbook's named styles.

## Why #170 is folded in

That last one *is* #170, and it's this page's Excel card rather than a separate surface — #170's own description says it depends on this page existing. `analyze_xlsx()` reads the named styles into `Analysis.styles_present`, the same field that means "the style names this document offers" for docx.

It answers a question nothing could answer before: which values a `styles:` directive may reference as `style:<Name>` — a set defined by a binary file on a volume, previously discoverable only by opening it in Excel. A test asserts the page and `load_template_styles()` agree.

**Excel is the odd slot and the page says so:** no default workbook ships, so removing a custom one leaves *no* named styles rather than restoring a fallback. Revert is worded and coloured differently there.

## Two things worth knowing for later

**Route ordering is load-bearing.** `/base/{slot}/download` also matches `/{kind}/{name}/download`, and Starlette takes the first route that fits — so the `/base/…` routes are declared ahead of the generic block. Losing that degrades to a silent redirect rather than an error, so `test_download_route_is_not_swallowed_by_the_generic_pattern` pins it.

**"Custom" is decided by path identity, not directory naming.** I first used `template_utils._classify_template_source()`, which looks for a path part called `custom_templates` — and the badge came out wrong in testing. It compares against the exact path a replacement is written to now. The heuristic also misreports a *default* as custom when the bundled defaults sit under such a directory, which would offer to revert a file the UI cannot delete; that's the case the test pins.

## Review round (c3115cb)

- **`base_revert` had a TOCTOU** — `is_file()` then an unconditional `unlink()`, so two tabs reverting at once meant the second raised `FileNotFoundError` into a 500. Restructured so `unlink()` decides, rather than bolting a catch onto the check-then-act.
- **The unusable-file gate was too narrow and copied three times.** `analyze_xlsx` has a second failure — the workbook opens but its styles don't read — whose message isn't `"Could not open"`, so it installed anyway. Replaced with `analysis.is_unusable()`, used by `draft`, `reupload` and `base_upload`.

## Verification

`ruff` clean, **2119 passed**. Mutation-tested:

| Mutation | Result |
|---|---|
| Back to the directory-name heuristic | 1 failed |
| Base routes registered after the generic ones | 3 failed |
| Upload skips the extension check | 1 failed |
| A rejected upload still writes the file | 2 failed |
| Revert offered with nothing to revert | 2 failed |
| Back to check-then-act on revert | 1 failed |
| Gate narrowed back to `"Could not open"` | 2 failed |
| Gate widened until any warning is fatal | 1 failed |

The first of those initially **passed** — my fixture names its directory `custom_templates`, so the heuristic worked there and the test wasn't pinning the improvement. Added a case where the two genuinely disagree.

The new page is also in `test_admin_assets.py`'s sweep (13 page types now), so the no-external-assets and `lang` invariants cover it.

## Docs

`docs/admin-ui.md` describes the page and the Excel exception. `docs/development/admin-ui.md` gains the module, both gotchas above, the `is_unusable` rule, and the test row. `AGENTS.md`'s layout map names `base_templates.py`.


dvejsada:
@claude review

A new page with destructive operations, so the areas I'd most want checked:

1. **Revert deletes a file.** `base_revert` unlinks `custom_templates/<fixed name>`. The filename comes from the slot table and never from the request, but please check the whole path: can any input reach `custom_path()`, and is there a state where revert could remove a *bundled default* rather than a custom file? Also whether the "no custom file to remove" branch can race with a concurrent revert.

2. **Route ordering.** `/base/{slot}/download` also matches `/{kind}/{name}/download`; I rely on declaration order and pin it with a test. Is that sound, or is there a shape that removes the ambiguity entirely rather than depending on ordering?

3. **`source_of()` vs `template_utils._classify_template_source()`.** I deliberately don't use the existing helper — it classifies by looking for a path part named `custom_templates`, which misreports a default as custom when defaults sit under such a directory. Is the exact-path comparison right, and does `.resolve()` introduce any problem (symlinked volume, missing file, permissions)?

4. **Upload validation ordering.** Extension, then size, then openability, then write. Is anything checked too late, and does a rejected upload leave the installed file genuinely untouched in every branch?

5. **#170 folded in.** Reading named styles via `workbook._named_styles` mirrors `xlsx_tools/styles.py`, but it's a private attribute in both places — worth flagging if there's a supported route.

---

dvejsada:
The TOCTOU was real and the tail turned up a second thing worth fixing. Both in c3115cb.

**1. `base_revert`.** Confirmed — `unlink()` on an already-gone file raises `FileNotFoundError`, so the friendly "nothing to remove" page becomes a 500 for whichever tab loses.

I didn't take the suggested shape. Wrapping `unlink()` in a `try` while keeping the `is_file()` check leaves the check-then-act in place and adds a second path that means the same thing. The route now lets `unlink()` decide — one syscall, no window:

```python
try:
    target.unlink()
except FileNotFoundError:
    return _base_page(..., message="There is no custom file to remove.")
```

Pinned by a test that makes `unlink` win the race and then raise, so it exercises the `except` branch rather than the check.

**2. The `analyze_xlsx` gap from your "not blocking" list.** You called it theoretical and it is — but the gate was also copied in three routes with three chances to drift, and the failure mode is installing a file whose entire purpose has failed while showing only a note. Replaced with `analysis.is_unusable()`, used by `draft`, `reupload` and `base_upload`, matching any warning that opens `"Could not …"`.

I checked that widening doesn't swallow advisories: the pptx layout/role/footer notes, the docx unbalanced-conditional note and the email encoding note all begin otherwise. The tests pin it from **both** sides — narrowing it back to `"Could not open"` fails, and so does widening it until any warning is fatal.

**3. You're right about my PR description.** I wrote "extension, then size"; the code checks size and emptiness first in `_read_upload`, then extension. Both precede any write so correctness is unaffected, but the description was wrong and I've corrected it.

**4. Two things I'm deliberately leaving.**

- **Route ordering → regex convertor.** Agreed it would make the ordering irrelevant rather than load-bearing, and it's the better shape if this class of route grows. For five fixed slots I'd be adding a constrained `{kind}` convertor that every existing route then depends on, to replace a dependency that a regression test already catches. Worth revisiting when a second such collision appears.
- **`/download` returning 200 with a not-found page.** You flagged it as pre-existing rather than introduced here, and that's exactly why I'm not changing it: every not-found in this UI renders a page, and making the download route the single one that 404s trades a consistent oddity for an inconsistent one. If it should change, it should change for `/{kind}/{name}/download` too, as its own thing.

Agreed on the rest — `custom_path()` taking no caller input, revert structurally unable to reach `default_templates/`, `.resolve()`'s `OSError` fallback, and `_named_styles` mirroring the existing precedent rather than adding a second divergent reader.

`ruff` clean, **2119 passed**. Three new mutations verified.

---

PR #179, opened 2026-09-19
admin: make the upload limit configurable, and check it before reading

### LINKED ISSUE #172: Admin UI: the 10 MB upload limit will reject the brand decks the feature is for
`admin/app.py:64`:

```python
# Reject uploads larger than this (read fully into memory before validation).
MAX_UPLOAD_BYTES = 10 * 1024 * 1024  # 10 MB
```

A designer's `.pptx` or `.potx` with a logo, a cover image, a few section backgrounds and an embedded font clears 10 MB routinely — and a corporate brand deck is precisely the file the PowerPoint template feature exists to accept. The limit is a reasonable default for a `.docx` letterhead and too low for the case it is most likely to meet.

The comment names the real constraint: the file is read fully into memory before anything validates it, in `draft` and `reupload` (`await upload.read()`), and again in `analyze_pptx`, which stages the bytes to a temp file to open them. So the limit is not arbitrary — raising it raises peak memory per concurrent upload.

Two options, worth deciding rather than just bumping the number:

1. **Per-kind limits.** Keep 10 MB for docx and email, raise pptx to something realistic (50 MB). Smallest change, fits the per-kind descriptor table the component work introduces.
2. **Stream to a temp file** and validate from the path. `analyze_pptx` already wants a path — it writes one itself — so streaming would remove a copy rather than add work. `write_asset` would move the temp file into place instead of taking bytes.

Either way, make the limit configurable (`ADMIN_MAX_UPLOAD_MB`), which per `AGENTS.md` means `config.py`, `.env.example` and `docs/configuration.md` together, enforced by `tests/test_config_docs.py`.

Also worth fixing alongside: the size check happens *after* the whole body is read, so an oversized upload costs the memory before it is refused. Starlette exposes the content length; checking it first makes the limit a real guard rather than a message.

dvejsada:
Closes #172.

The cap was a hard-coded 10 MB, which a designer's brand deck clears routinely — exactly the file the PowerPoint template feature and the base-templates page from #178 exist to accept. It's now `ADMIN_MAX_UPLOAD_MB`, defaulting to **25**.

## The more interesting half

The issue also said the size was checked *after* the whole body had been read, so an oversized upload cost the memory before being refused. Investigating that turned up both why it matters less than it looks and how to fix it properly:

**Starlette already spools a multipart upload to a temp file above 1 MB** (`MultiPartParser.spool_max_size`). So the request itself was never the expensive part — `await upload.read()` was, because it pulls that spooled file into a `bytes`.

`_read_upload()` now reads `UploadFile.size` **first** (with a seek fallback for a file object that arrived another way) and refuses without reading. The post-read check stays as a backstop for an upload whose size can't be known in advance. A test patches `UploadFile.read` and asserts an oversized body is never read at all.

## Two things I decided against

**Per-kind limits**, which the issue floated as its first option. Memory is bounded by the largest *accepted* upload whatever its kind, the extension check already rejects a wrong-format file, and five limits are five things to misconfigure. One documented number is easier to reason about — happy to reverse this if you'd rather keep `.html` wrappers on a tight leash.

**Streaming an accepted upload** straight to its destination. That needs the analysers and `store.write_asset()` to take a file object rather than bytes, which is a wider change than this issue warrants. Recorded in `docs/development/admin-ui.md` as what to do if the limit is ever raised far — an accepted 25 MB upload still costs 25 MB.

## Tests

`tests/test_admin_app.py` now runs against `ADMIN_MAX_UPLOAD_MB=1`, so the oversized cases cost a megabyte instead of the default 25 **and** prove the variable is live end to end — the refusal message quotes the configured limit. Config-level tests cover the default, an override, and every unusable value (`""`, `"0"`, `"-5"`, `"nonsense"`, `"12.5"`) falling back rather than disabling the limit or crashing startup.

## Verification

`ruff` clean, **2128 passed**. Mutation-tested:

| Mutation | Result |
|---|---|
| Drop the pre-read size check | 1 failed |
| Hard-code the limit, ignoring config | 3 failed |
| Let a bad env value disable the limit | 2 failed |
| Undocument the variable | 1 failed |

## Config

`config.py`, `.env.example` and `docs/configuration.md` together, as `tests/test_config_docs.py` requires.

---


dvejsada:
@claude review

Small diff, but it touches every upload path in the admin UI. Worth checking:

1. **`_upload_size()`.** It trusts `UploadFile.size` when present. Is that value authoritative for what `read()` will return, or can they disagree (a truncated multipart part, a chunked request, a client-supplied length)? If they can, the pre-read check could refuse a valid file or admit an oversized one — the post-read backstop covers the second case but not the first.

2. **The seek fallback** does `seek(0, 2)` / `tell()` / `seek(0)` on `upload.file`. Does that leave every stream this code can actually receive positioned correctly for the subsequent `read()`, including Starlette's `SpooledTemporaryFile` after it has rolled to disk?

3. **Two decisions I made against the issue's own suggestions** — skipping per-kind limits, and not streaming accepted uploads to disk. Please push back if either is wrong; the reasoning is in the PR body.

4. **Config fallback.** Any `ADMIN_MAX_UPLOAD_MB` value that would disable the limit, overflow, or crash startup rather than falling back to 25?

5. I removed the `MAX_UPLOAD_BYTES` module constant in favour of `ctx.max_upload_bytes`. Anything still expecting the old name, in code or docs?

---

dvejsada:
Correct on all counts — the dead constant is gone in 9e446bd.

**The finding.** `DEFAULT_MAX_UPLOAD_BYTES` matched exactly once in the repo: its own definition. I added it as a "fallback ceiling when no config is to hand", and you're right that the case can't arise — `ctx.max_upload_bytes` always reads `config.admin.max_upload_bytes`, and the pydantic field carries its own default. Deleted rather than wired into a guard for a state that doesn't exist; a name that looks load-bearing and isn't is worse than no name.

**Q1 is the useful answer here.** I asked whether `.size` could disagree with what `read()` returns because I couldn't tell from the code I was touching. Your detail settles it — Starlette computes `.size` while writing each chunk to the spooled file, so it counts bytes that actually arrived rather than a client-declared length, and a truncated part fails in `req.form()` before `_read_upload()` runs. That's precisely the fact that makes trusting it instead of reading first safe, so I've put it in the docstring rather than leaving the next reader to re-derive it.

**Q2** — noted on both the rollover behaviour and the sync-call point. Agreed `seek`/`tell` aren't worth offloading; they're metadata operations, and the read that follows is the expensive part and already bounded.

**Q4's `"1_000"` note** is a real quirk of `int()` and I'm leaving it. Someone writing `ADMIN_MAX_UPLOAD_MB=1_000` gets 1000 MB, which is what they meant — stricter parsing would reject a value whose intent is unambiguous, for no safety gain, since a large limit is a choice rather than a failure.

**Q3** — thanks for pushing on both deferrals rather than just accepting the PR's framing.

`ruff` clean, **2128 passed**.

---

PR #180, opened 2026-09-19
admin: surface what a build worked around on the Status page

### LINKED ISSUE #173: Status page: per-template warning counts from the warnings channel
`warning_channel.py` exists so that a build which skips, substitutes or degrades something the caller asked for says so. Per `AGENTS.md`, Word, Excel and PowerPoint all carry a channel and every code has a severity — `error` (not in the file), `warning` (there but not as asked), `info` (a substitution the caller will not mind).

The Status page sees none of it. `metrics.ToolStat` (`metrics.py:32`) holds `calls`, `errors`, `last_called` and `last_error`, and `record_call` / `record_error` are the only recorders. So a template that renders successfully every time while silently falling back to a default style, dropping a table row or laying a slide out on the wrong layout is indistinguishable on screen from one that is working perfectly.

That is the gap worth closing, because the warnings channel was built on the premise that the caller gets a success response and never reads the server log — and the Status page is where an operator would go to find out what has been degrading.

Add `warnings` and `last_warnings` to `ToolStat`, a `record_warnings(kind, name, warnings)` alongside `record_error`, and a Warnings column in the usage table. Group by severity so `info` substitutions do not drown the `error`-severity ones that mean content is missing from the file.

The call sites are where `main.py` attaches warnings with `_with_warnings()`, which is already the single point every channel-carrying tool passes through — so this should be one recorder call, not one per tool.

Static tools are worth counting too, not only template tools: the usage table is currently template-only, but a Word conversion degrading is the same operator question.

dvejsada:
Closes #173.

`warning_channel.py` exists because a build that substitutes a style, drops a table row or lays a slide out on a fallback layout still returns a **success** — the caller never questions it, and the detail must not live only in the server log.

But `metrics.ToolStat` counted calls and errors only. So on the Status page that degrading build looked exactly like a clean one: the one place the product records "this worked, but not as asked" was the one place the operator view couldn't see.

## What it looks like now

```
Tool                           | Kind | Calls | Errors | Warnings | Most recent problem
create_powerpoint_presentation | pptx |     1 |      1 | —        | layout 'kpi' not found
create_excel_document          | xlsx |     1 |      0 | 1 info   | substituted Calibri for Brand Sans
create_word_document           | docx |     3 |      0 | 3 error  | line 12: table row 2 was dropped

Warnings reported: 3
```

That Word row is the case the issue was about — three successful calls, zero errors, and three rows silently missing from the output.

## Severities are kept apart, not totalled

`info` is a substitution the caller won't mind; `error` means something they asked for **isn't in the file**. Lumping them together lets a noisy-but-harmless font swap mask a real loss, so the column splits them and the tile counts `ToolStat.degraded` — everything except `info`. Note the tile above reads 3, not 4: the Excel `info` is counted but not chased.

A severity the table doesn't recognise counts as *degraded* rather than being dropped, so a new code can't go quiet by being unrecognised.

## Where it records, and why that matters

From `main._with_warnings()` — the single point every channel-carrying tool passes through on success. It now takes the tool's `kind` and `name` as **required** arguments, so a new tool that forgets them fails there instead of going silently uncounted. A test walks `main.py`'s call sites and asserts each passes an identity.

That also puts static tools in the usage table for the first time. A Word conversion quietly degrading is the same operator question as a template tool doing it, so it's **"Tool usage"** now rather than "Template usage", and the static handlers record their errors too.

## The card

The counters say a tool is degrading; a second card says *how*, listing what recent builds actually worked around with the location the channel captured (`line 12: …`, `sheet Data, cell B2: …`). Without it the only way to find out is the server log — exactly where the warnings channel exists to stop this information living.

## Verification

`ruff` clean, **2146 passed** (18 new). Mutation-tested:

| Mutation | Result |
|---|---|
| Never record warnings | 11 failed |
| Lump all severities into one | 8 failed |
| Count `info` as degraded | 1 failed |
| Remove the recorder from `_with_warnings` | 2 failed |
| Drop the card from the page | 2 failed |

## Docs

`shared-modules.md` and `architecture.md` on what `metrics` now tracks and why `_with_warnings` takes an identity; `AGENTS.md`'s warnings rule; the admin dev page's test table.

---


dvejsada:
@claude review

This one changes a signature in `main.py`'s hot path, so:

1. **`_with_warnings` now takes required `kind`/`name`.** Please check every call site got them and that nothing else in the repo (or docs) calls it with the old signature. The requiredness is deliberate — I want a new tool to fail loudly rather than go uncounted — but say so if that's the wrong trade for a function on the success path of every document build.

2. **Recording happens inside `_with_warnings`, before the early return** for a clean build. Is that the right place, given it now has a side effect the name doesn't advertise? I considered a separate explicit call per handler and rejected it as three chances to forget; push back if the coupling is worse than the risk.

3. **`metrics.record_warnings()` accepts anything iterable with a `severity`** — a `WarningChannel`, its `records()`, or PowerPoint's own list. Is that duck-typing too loose? Note `WarningChannel.__iter__` yields `records()`, which *includes* the synthetic `warnings_truncated` entry, so a truncated build counts that as an `info` warning. Intended, but worth a second opinion.

4. **Thread safety.** `record_warnings` takes `_LOCK` like the other recorders, but builds run concurrently on `run_blocking`'s pool — is the mutation of `warnings_by_severity` and `last_warnings` safe as written?

5. **`ToolStat.degraded` treats an unknown severity as degraded.** I think failing loud is right for a code whose severity table is wrong, but it's a judgement call.

---

dvejsada:
The regex fragility was real and worse than "works today". Fixed in 90f778b.

**The guard could stop enforcing silently.** Confirmed your reading — capture 3 truncates at `len(slides`:

```
'result, warnings, "pptx",\n    "create_powerpoint_presentation",\n    slide_count=len(slides'
```

I went further and demonstrated it rather than reasoning about it. Changing the Word site's kind to `str(len("docx"))` — a non-literal identity, exactly what the test exists to reject — gives:

```
OLD regex guard verdict: PASSES (missed it)
  truncated capture: 'result, warnings, str(len("docx"'
NEW ast guard: 1 failed
```

So the old test passed a mutation that removed the very thing it checks. For a test whose only job is to stop a future call site going uncounted, that's a failure of the test rather than a nit.

It parses with `ast` now and checks each call has ≥4 positional args with the third and fourth being non-empty string literals. Three mutations fail it: a non-literal identity, a missing one, an empty one.

**Both minor observations taken:**

- `_warnings_card(ctx)` dropped the unused parameter.
- The tile's re-filtering meant "what counts as degraded" lived in two places. `metrics.degraded_total()` answers it now, built on `ToolStat.degraded`. That also retired `counts_by_severity()` — once the tile stopped filtering, nothing but the tests called it, and a public function only its own tests use is the dead-code shape flagged on #179.

**On Q4**, your distinction is the right one and I'd not stated it precisely: the mutations are individually lock-guarded, and the non-atomic gap is between `record_call` and `record_warnings`, so a reader can see a call counted before its warnings. That's eventual consistency on a counter read by a human refreshing a page, not corruption — leaving it.

**Q1, Q2, Q3, Q5** — agreed, no changes. Useful that you checked `SlideWarning.severity` actually exists rather than taking the duck-typing claim on trust, and confirmed `WarningChannel.__iter__` delegates to `records()` so the truncation notice is counted deliberately.

`ruff` clean, **2146 passed**.

---

Issue #151, opened 2026-09-19
Admin UI status page: surface warnings, and make the log usable


dvejsada:
The Status page reports calls, errors, uptime and a tail of the log. Since `warning_channel.py` landed, that is no longer the whole picture: Word, Excel and PowerPoint all report what a build worked around, and the Status page cannot see any of it. A template that renders but silently substitutes a missing style, drops a table row or lays a slide out on a fallback layout looks identical to one that is working perfectly.

`metrics.py` tracks only `calls`, `errors`, `last_called` and `last_error` (`ToolStat`, `metrics.py:32`). The warnings never reach it.

The log view has the complementary problem: 150 lines, one two-state filter (info / warnings+errors), no search, no per-logger filter, and a manual "Refresh" link — so it is only usable if the thing you want is in the last 150 records and you can spot it by eye.

Sub-issues below.

PR #181, opened 2026-09-19
admin: make the Status page's log view searchable

### LINKED ISSUE #174: Status page: the log view needs search, a logger filter and auto-refresh
The recent-activity view is usable only if what you want is in the last 150 records and you can spot it by eye.

- `metrics.install_log_capture` keeps a 300-record ring buffer; `_status_view` renders `recent_logs(min_level, limit=150)`.
- The only filter is two states — All (INFO and above) or Warnings & errors (WARNING and above) — as two links at `app.py:929`.
- No text search, so finding every record mentioning one template means scrolling.
- No per-logger filter, though each record already carries `logger` and it is rendered as its own column.
- "Refresh" is a manual link (`app.py:935`). Diagnosing a failing tool call means clicking it after every attempt.

Four additions, each small and independent:

1. **Text search** over the message, as a query parameter alongside `level`.
2. **Logger filter** — a dropdown built from the distinct `logger` values in the buffer, which makes "just the template store" or "just the upload backend" one click.
3. **Full level choice** rather than two states: DEBUG / INFO / WARNING / ERROR. `recent_logs` already takes a `min_level`, so this is the control catching up with the function.
4. **Auto-refresh**, off by default, as a meta refresh or a few lines of inline JS re-fetching the log block. No external library — that constraint is deliberate and tested.

Worth revisiting the 300-record capacity at the same time: with DEBUG enabled it holds well under a minute of traffic, and the buffer is in-process and bounded, so a larger default costs little.

dvejsada:
Closes #174 — and with it #151, whose other half was #173.

The log view showed 150 records behind a two-state All / Warnings toggle and a manual Refresh link, so it was usable only if what you wanted was near the top and you could spot it by eye. #173 made the page say *that* a tool is degrading; finding the surrounding detail still meant scrolling.

## What it does now

Filters by **level**, by **which part of the server logged it**, and by a **case-insensitive search** over the message *and* the logger name — so `store` finds both a message mentioning it and everything `admin.store` logged. Auto-refresh at 10/30/60s, **off by default**, because a page that reloads while you're reading it is worse than one you refresh yourself.

Every filter is a query parameter, so a filtered view is a URL you can bookmark or paste to someone else.

## Three things worth a look

**Filtering happens before the limit.** `metrics.recent_logs()` applies level/source/search over the whole buffer and *then* takes the newest N, so the limit bounds **matches** rather than records scanned. Limiting first silently loses a match sitting behind the newest page.

That one is also a lesson in test design: my first version of the test logged its needle **last**, so it fell inside the newest page either way and the mutation passed. It now buries the needle behind twice the limit, which is the only arrangement that distinguishes the two.

**Levels that can't match aren't offered.** The buffer captures at the server's configured level, so on an INFO server "debug and above" is a filter that can only ever come back empty and reads as broken. `available_levels()` withholds it and the page says to set `DEBUG=true` instead.

**`error` now means errors.** It previously meant "warnings and errors", so a bookmarked `?level=error` shows fewer lines than before. The `warning` level carries the old meaning. Flagging it as the one behaviour change here.

## Smaller calls

- **Buffer 300 → 1000.** At DEBUG, 300 held well under a minute of traffic; it's in-process and bounded, so a longer window costs a few hundred KB. I did **not** make it configurable — one more variable to misconfigure for something with no failure mode.
- **Auto-refresh is an inline timer**, not a `<meta http-equiv="refresh">` (it keeps the query string) and not anything external (the UI loads no third-party scripts, per `AGENTS.md`).

## Review follow-up (`1b2d6e7`)

**The log scan now snapshots the ring buffer before filtering it.** `recent_logs()` iterated `_LOG_HANDLER.records` directly while builds on worker threads append to it; a `deque` raises `RuntimeError: deque mutated during iteration` when that lands mid-scan, so a build finishing while someone had the Status page open could 500 it. `counts_by_level()` already snapshotted — `recent_logs()` now matches. The gap predates this PR, but 1000 records and a per-record substring match widen the window it can land in.

Its regression test is deterministic rather than a thread race: a record whose `__getitem__` appends to the deque puts the hand-off exactly where a real one would.

**`?refresh=` is clamped to `REFRESH_CHOICES`.** `max(0, int(refresh))` accepted `?refresh=999999`, which armed a timer that no `<option>` renders as selected — the control read "off" while the page reloaded. Parsing moved into `admin/views/status.refresh_seconds()`, beside the choices it clamps against and next to `level_no()`, leaving the route a pass-through.

## Verification

`ruff` clean, **2171 passed** (25 new). Mutation-tested:

| Mutation | Result |
|---|---|
| Ignore the search | 4 failed |
| Ignore the source filter | 1 failed |
| Make the search case-sensitive | 4 failed |
| Limit before filtering | 1 failed |
| Offer an unreachable level | 1 failed |
| Don't sanitise `refresh` | 1 failed |
| Force auto-refresh on | 6 failed |
| Filter the live deque instead of a snapshot | 1 failed (`RuntimeError`) |
| Clamp `refresh` with `max(0, …)` instead of the choices | 2 failed |

## Docs

`docs/admin-ui.md` describes the filters; `docs/development/admin-ui.md` and `shared-modules.md` record why filtering precedes the limit, why levels are withheld, why the scan snapshots the buffer, and why the refresh interval is clamped.

---


dvejsada:
@claude review

Areas I'd most want checked:

1. **`recent_logs()` filters the whole buffer on every request.** With the capacity raised to 1000 and a search over message + logger name, that's up to 1000 lowercase substring checks per page load, under `_LOG_HANDLER`'s records while other threads are appending. Is iterating the deque while it may be mutated safe here, and is the cost acceptable for a page a human loads?

2. **`available_levels()` reads `_LOG_HANDLER.level`** via `metrics.capture_level()`. Is that the right source of truth — could the handler's level and the root logger's level disagree in a way that makes the page withhold a level that *would* match, or offer one that can't?

3. **The `?level=error` semantic change.** It used to mean "warnings and errors". Anything else in the repo or docs that assumes the old meaning?

4. **`components.auto_refresh()` emits `setTimeout(...location.reload())`.** It interpolates an int into inline JS — check that path for anything that could reach it unsanitised, and whether reloading discards in-progress form state anywhere on this page.

5. **The source filter matches by top-level package prefix.** Is `record_logger == source or startswith(source + ".")` right, or are there logger names in this codebase where that groups badly?

---

dvejsada:
Acted on the review in `1b2d6e7`.

**Taken — the deque finding (Q1).** Confirmed before fixing: `deque` does raise `RuntimeError: deque mutated during iteration` on an append mid-loop, including when the buffer is not yet at `maxlen`. `recent_logs()` now snapshots with `list(...)` first, matching `counts_by_level()`.

The regression test is deterministic rather than a thread race — it puts the append exactly where the hand-off would land, inside the first field the scan reads off a record:

```python
class _AppendsWhenRead(dict):
    def __getitem__(self, key):
        handler.records.append(dict(self))
        return super().__getitem__(key)
```

Mutation-tested: with the `list(...)` removed it fails with `RuntimeError` at `metrics.py:244`; restored, it passes. It also asserts something was captured first, so it can't pass vacuously on an empty buffer — the first version of it did exactly that, because the root logger sits at WARNING under pytest and the INFO probe never reached the handler.

Agreed this is pre-existing rather than a regression, and that the scan cost itself is not a concern.

**Taken — both doc misses (Q3).** `shared-modules.md` now says `LOG_BUFFER_CAPACITY` rather than a number that can drift again; the `admin-ui.md` Status bullet drops the trailing "filterable to warnings/errors" left over from the two-state toggle. Also recorded both new invariants in `development/admin-ui.md` — the snapshot rule, and the clamp below.

**Taken — the `?refresh=999999` note (Q4).** You called it harmless, and it is, but the mismatch it creates isn't: the page reloads on a timer while the `<select>` renders no option as chosen, so the control reads "off". `refresh_seconds()` now clamps to `REFRESH_CHOICES` and lives beside them in `admin/views/status.py`, next to `level_no()`, leaving the route a pass-through. The existing "unusable value" test gained `999999` and `45`; both fail against the old `max(0, int(refresh))`.

**Not changed — Q2 and Q5.** Nothing to do: both check out, and the `install_log_capture()` caveat in Q2 is a note for whoever changes that call, not a change to make now.

2171 tests pass, `ruff` clean.

---

Issue #167, opened 2026-09-19
Admin UI: master-YAML templates are listed read-only with no way to inspect or adopt them


dvejsada:
`_template_table` (`admin/app.py:496`) lists templates that are live but not UI-managed as a bare row: the name, a dash, "● Live", and the badge `from master YAML — read-only`. That is the entire affordance. You cannot see the template's description, its arguments, its style mapping, or which file it uses — the UI knows all of it, since `gather_specs` returned the spec to compute the row, and shows none of it.

For anyone whose templates predate the admin UI, that is most of their templates, permanently second-class.

Two steps, useful separately:

**1. Inspect.** Make the name link to a read-only detail page: description, arguments table, asset filename, style mapping, and the same "what we found in the document" analysis the edit page runs. No form, no save. This alone removes most of the frustration and is nearly free — it is the existing `_analysis_report` plus a non-editable rendering of the spec.

**2. Adopt.** An "Adopt" action that copies the master-YAML spec into the `*.d` merge layer, after which it edits like any other UI template. The merge order means the adopted copy wins, so the master entry can stay where it is and the admin's hand-written YAML is never modified — consistent with the store's stated rule.

Adopt needs care where the master entry points at a file in `default_templates/` rather than `custom_templates/`: the managed spec must either keep pointing there or copy the asset into the writable directory. Copying is probably right, since a template you can edit but whose document you cannot replace is a confusing half-state.

PR #182, opened 2026-09-20
admin: disable a template without deleting it, and rename one

### LINKED ISSUE #165: Admin UI: rename a template, and disable one without deleting it
**Delete is the only off switch.** To stop the AI reaching for a template — a seasonal one, a broken one, one being revised, one whose legal text is under review — the only supported move is to destroy its configuration: the spec file goes, arguments and descriptions and all, and the source file is orphaned on disk where nothing lists it.

There is also no rename. The name field is explicitly disabled on the edit form (`app.py:654`, `app.py:769`) and only submitted as a hidden value, so a typo in a tool name is permanent — or rather, fixable only by deleting and re-entering everything.

**Disable.** Add an `enabled: false` key to the spec and an Enable/Disable control on each row. Disabled means: keep the spec file, keep the asset, but unregister the live tool (`unregister_docx_template` / `unregister_email_template`; for pptx, exclude it from the registry so it stops being a value for the `template` argument). The loaders need to skip disabled specs at startup too, so the state survives a restart. Show it in the existing Status column, which already distinguishes Live from Not live.

**Rename.** Write the spec under the new name, unregister the old tool, register the new one, delete the old spec file. Decide whether the asset is renamed alongside: renaming it keeps `custom_templates/` legible but breaks any master-YAML entry that points at the same file, so the safe default is to leave the asset filename alone and let the spec keep pointing at it.

Both need `validate_name` on the target and a collision check against existing specs.

dvejsada:
Closes #165.

Deleting was the only off switch. To stop the AI reaching for a template — a seasonal one, one being revised, one whose legal text is under review — you had to destroy its configuration: the spec file went, arguments and descriptions with it, and the source file was orphaned on disk where nothing listed it. There was no rename either: the name field was a disabled input, so a typo in a tool name was permanent short of deleting the template and re-entering everything.

| | before | after |
|---|---|---|
| Live | Edit · Delete | Edit · **Disable** · Delete |
| Disabled | — | Edit · **Enable** · Delete |

## Where disabling lives

`enabled: false` goes in the spec, and `gather_specs()` filters it — the one merge point the docx loader, the email loader and the PowerPoint registry all go through. Filtering **there** rather than at each registration site is the whole design: it is what makes the state survive a restart with nothing having to remember it, and what stops a future consumer of `gather_specs()` registering a disabled template by forgetting to check. Absent means enabled, so every spec written before this stays live — there is a test for exactly that upgrade path.

`is_enabled()` also accepts the string forms PyYAML leaves uncoerced. Only the full words become bools, so `enabled: n` and `enabled: disable` arrive as strings — and reading either as *enabled* would invert the author's intent silently, which matters most for `disable`, since it mirrors the UI's own button. A string in neither vocabulary is treated as enabled (the safe direction — a template the AI cannot call looks like a broken server) but logged with the template's name.

## Three traps worth a look

**A save was switching it back on.** The edit form carries no `enabled` control, so `build_spec()` rebuilds the spec without one — edit a disabled template and it came back live. The save route now reads the stored flag and carries it forward, and `AdminContext.sync()` is the single place that decides whether saving registers a template or takes it off.

**A create was reading as an edit.** `is_new` was dead — both `edit_form` callers hardcoded `False` — so the form you land on straight after uploading a new file rendered `original_name` too. A fresh upload whose name collided with an existing template inherited its `enabled` flag and overwrote its args and description in silence. `configure_page` now says `is_new=True`, the route keys off whether `original_name` is *present* rather than whether a stored spec turns up, and a create landing on an occupied name is refused with the message a rename gets.

**PowerPoint's fallback had to learn a distinction.** With nothing configured, the registry synthesises `16_9`/`4_3` from the historical filename slots. Disabling every template is not "nothing configured" — it is a choice, and falling back there would hand the built-ins straight back under names a deck would still build on, silently, off the wrong design. The fallback now separates "nothing readable" from "all turned off", and still runs for the malformed-config case it exists for. I mutation-tested **both** directions, because a guard like this fails quietly in whichever one you don't check.

## Rename

Goes through `store.rename_spec()`, which owns the collision check, the name validation and the write-before-unlink ordering — so an interrupted rename leaves two templates rather than none. A failed unlink raises `OSError`, which the route catches alongside `TemplateStoreError`. The asset keeps its own filename: renaming it would break any master-YAML entry pointing at the same file, and the spec names it explicitly anyway.

**Correction to an earlier version of this description.** It said the safe ordering was what ran. It wasn't: the route reimplemented save-then-delete inline and never called `rename_spec()`, so the method the test covered was the one production didn't use — and the "unlinks before it writes" mutation in the table below passed against dead code. `17d8add` fixes it. The check that matters now: making `rename_spec()` a no-op fails the *route-level* rename tests, not just the direct one.

## What I decided against

- **Renaming the asset alongside the template.** It would keep `custom_templates/` legible, but silently breaks any master-YAML entry pointing at the same file. The issue called this out and I agree with its default.
- **A separate rename page.** Rename is a property of the template, and the edit form already had the field — disabled. Re-enabling it is smaller than a second flow, and a typo is fixed where you notice it.
- **A collision check against master-YAML names.** Renaming onto a master name creates a `.d` spec that overrides it — exactly what creating a new template with that name does, and the documented adopt behaviour (#167). A special case only here would make the two paths disagree.
- **Verifying `original_name`'s provenance.** A forged value would delete an unrelated template as a side effect of saving. Real, but not a privilege boundary: anyone who can forge it is past the admin password and can call the delete route directly. A signed token or an asset-filename match doesn't belong in this PR.

## Verification

`ruff` clean, **2218 passed** (47 new). Every one of these mutations fails the suite:

| Mutation | Result |
|---|---|
| `gather_specs()` does not filter disabled specs | 3 failed |
| `is_enabled()` ignores the string forms hand-edited YAML produces | 5 failed |
| An unrecognised `enabled` value is accepted in silence | 1 failed |
| Save registers instead of syncing | 1 failed |
| Save drops the carried-forward `enabled` flag | 2 failed |
| Carry-forward keyed on the stored spec, not on `original_name` | 2 failed |
| A create silently overwrites an existing template | 2 failed |
| The create form renders `original_name` again | 1 failed |
| `rename_spec()` is a no-op | 4 failed |
| A failed unlink 500s instead of reporting | 1 failed |
| Rename leaves the old spec behind | 2 failed |
| Rename leaves the old tool registered | 1 failed |
| Rename clobbers an existing template | 1 failed |
| The name field goes back to display-only | 1 failed |
| The pptx registry ignores `enabled` | 2 failed |
| Disabling every pptx template resurrects the built-ins | 1 failed |
| The legacy fallback never runs at all | 1 failed |

The restart test carries a positive control — it asserts the loader finds the template *before* disabling it, so a loader pointed at the wrong config could not make the real assertion vacuous.

## Docs

`admin-ui.md` (Disable/Enable/Rename, and Delete now points at Disable for the "I just want it off" case), `templates.md` (the key in the YAML example), `architecture.md`, `dynamic-templates.md`, `shared-modules.md`, `powerpoint.md` and `development/admin-ui.md` (the invariants — including that rename goes through the store, and what `original_name` separates), plus an `AGENTS.md` rule that the filtering lives in `gather_specs()` and never at a call site.

---


dvejsada:
@claude review

Areas I'd most want checked:

1. **`gather_specs()` now filters by default.** Every caller gets the new behaviour unless it opts out. I changed `admin/app.py`'s `_master_specs()` and the pptx registry to pass `include_disabled=True` where they need the unfiltered view — is there a caller I missed, or a place where silently dropping a spec produces a worse failure than registering it would?

2. **The save route's enabled carry-forward.** It reads `form["original_name"]`, falls back to `spec["name"]`, and looks the stored spec up from that. Is there a submission — a new template, a rename, a stale form, a tampered `original_name` — where it reads the wrong spec's flag, or re-enables something it shouldn't?

3. **Rename ordering.** `save_spec(new)` → `delete_spec(old)` → `unregister(old)`. If `delete_spec` fails you keep two specs and two live tools pointing at the same asset. Is that the right failure, and is there an ordering that fails better?

4. **`is_enabled()` accepts string forms** (`"false"`, `"no"`, `"off"`, `""`) because these files are hand-editable YAML. Is that helpful or is it a footgun — e.g. does any real YAML value round-trip into a string that now means the opposite of what the author intended?

5. **The pptx legacy fallback.** `all_disabled = bool(all_entries) and not entries`. Does that hold when the master YAML and the `.d` dir disagree, or when an entry is malformed *and* another is disabled?

---

dvejsada:
All three findings were real. Verified each against the code before touching anything, and fixed all three in `17d8add`.

**#3 is the one I should own.** `rename_spec()` was never called outside tests — `grep -rn "rename_spec" --include=*.py . | grep -v tests/` returns only its own definition. So the store had the safe ordering, my test proved the store had it, and the route an admin actually hits reimplemented the same sequence inline. Two implementations of one invariant, and the tested one was the unused one. My PR body claimed the safe ordering was what runs; that was wrong, and the mutation I ran ("`rename_spec()` unlinks before it writes") passed against a method production never called.

The route now calls `store.rename_spec()`. The check that matters: making `rename_spec()` a no-op now fails `test_renaming_moves_the_tool` and `test_renaming_a_pptx_template_moves_it_in_the_registry` — the route-level tests — not just the direct store one.

You were also right that the inline version wasn't equivalent. `delete_spec()` raises a plain `OSError` on a failed unlink and the route caught only `TemplateStoreError`, so the "two specs, two live tools" outcome I described in the PR body was really an unhandled 500 with the new spec written but never registered. Now caught, with a test that asserts the failure surfaces on `save_failed_page` rather than 500ing.

**#2 confirmed, and it was worse than the enabled leak.** `is_new` was dead — both `edit_form` callers hardcoded `False`, so it was never `True` anywhere. `configure_page` now passes `is_new=True`, which is true: it's the form you land on straight after uploading a new file.

The carry-forward now keys off whether `original_name` is *present*, not off whether a stored spec turns up. And a create landing on an occupied name is refused with the same message a rename gets — the silent overwrite predates this PR, but the distinction I needed for the `enabled` leak made the guard two lines, so closing it here rather than leaving a known hole seemed better than the alternative.

On the tampered-`original_name` point: agreed it's real, and I've left it. Anyone who can forge that field is already authenticated past the admin password and can call `/{kind}/{name}/delete` directly, so it isn't a privilege boundary — verifying provenance would mean a signed token or an asset-filename match, and neither belongs in this PR.

**#4 confirmed empirically.** `yaml.safe_load` leaves `n`, `N`, `y`, `disable` and `disabled` as strings; only the full words coerce. So the two values closest to the UI's own button label meant the opposite of what the author typed. All now recognised, with an `_ENABLED_WORDS` vocabulary beside them so a deliberate `yes` doesn't get warned about, and anything in neither logged with the template's name. Enabled stays the safe default — a template the AI cannot call looks like a broken server.

**#1 and #5** — thanks for tracing the callers and the `all_disabled` scenarios rather than eyeballing them. Nothing to change.

Seven new mutations, all caught: dropping the create-collision guard; the create form rendering `original_name` again; keying the carry-forward on the stored spec; not catching `OSError`; removing the new disabled words; dropping the warning; and `rename_spec()` as a no-op.

`ruff` clean, **2218 passed** (12 more than before).

---

PR #183, opened 2026-09-20
admin: a maintenance view of the files in custom_templates/

### LINKED ISSUE #166: Admin UI: a view of the files in custom_templates/ that no template points at

dvejsada:
Closes #166.

The uploads directory was write-only from the admin's point of view: files went in and were never shown again. Deleting a template keeps its source file by design, a rename or a replacement can strand one, and a file copied onto the volume by hand was equally unlisted — so orphans accumulated where nothing reported them and nothing could remove them.

The new **Source files** page lists every file with its size and mtime and what still points at it, orphans first because they are the only rows with anything to do. Only an orphan offers Delete, one at a time, with the filename on its own confirmation page.

```
leftover.docx              36 KB  2026-09-20 05:39   Unreferenced                          [Delete]
custom_docx_template.docx  36 KB  2026-09-20 05:39   Base template — Word                     —
handwritten.docx           36 KB  2026-09-20 05:39   Word template 'legacy' (master YAML)     —
letter.docx                36 KB  2026-09-20 05:39   Word template 'letter'                   —
```

## The only thing that really matters here

This page offers to delete an orphan, so **a file wrongly called an orphan is a file wrongly offered for deletion**. The reference map has to be complete, and three things count — each easy to miss:

- a **managed spec**;
- a **master-YAML entry**, read through the same merge the registry uses with `include_disabled=True`, because a disabled template still owns its file;
- a **base-template slot**.

The third is the one worth catching. Base templates live in the same flat directory under fixed names and *no spec names them*, so a scan built on specs alone calls `custom_docx_template.docx` an orphan and offers to delete the file that styles every Word document the server produces. The issue warned about this for master-YAML templates; the base slots are the same trap one layer over, and they arrived after the issue was written.

## Deleting is keyed off the scan, never the URL

The filename has to match one `scan()` produced — a basename read from the directory, so a crafted path matches nothing — and it has to *still* be an orphan when the request arrives, so a file that gained a reference since the page was rendered is no longer deletable. That also means no separate path validator: `validate_asset_filename()` is kind-specific (it checks the extension) and this page has to list a stray `.txt` too.

Routes are registered **before** the generic `/{kind}/{name}/…` ones, or `/files/{name}/delete` would be matched as a template delete. There's a test pinning that.

## The second-order hazard in the issue — reproduced, and narrower than described

The issue notes that `save_spec()` derives the asset filename from the template name and only raises when the asset is *missing*, so creating a template whose name matches a deleted one "silently adopts whatever stale file is still sitting there". I tested it:

- **Through the UI: not reachable.** The draft step writes the upload to `<name><ext>` first, so the stale file is *overwritten*, not adopted. I confirmed the document content is the newly uploaded one.
- **By POSTing `/{kind}/save` directly with an empty `asset_filename`: reachable.** `save_spec()` derives `<name>.docx`, finds the stale file, and adopts it with no indication.

I've left that alone rather than widen this PR — it's a save-route change, not a view, and this page at least makes the reuse visible afterwards. Happy to fix it in a follow-up; say the word and I'll add the guard.

## Verification

`ruff` clean, **2240 passed** (22 new). Each of these mutations fails the suite:

| Mutation | Result |
|---|---|
| Drop master-YAML entries from the reference map | 1 failed |
| Drop the base-template slots | 2 failed |
| Delete trusts the URL instead of the scan | 3 failed |
| Orphans not sorted first | 1 failed |
| Offer Delete on every row | 1 failed |

Plus tests that a disabled template still owns its file, that a shared file lists both templates, that a crafted path (`../config/docx_templates.yaml` and two encodings of it) deletes nothing, that a CSRF-less POST deletes nothing, and the end-to-end story: delete a template keeping its file, and the file shows up here as unreferenced.

## Docs

`admin-ui.md` describes the page. `development/admin-ui.md` gains the module, the completeness invariant and the delete-guard reasoning, and **loses** the "a kept source file is still invisible" limitation this closes. The delete page's own hint no longer claims kept files are listed nowhere — they are listed here now.

---


dvejsada:
@claude review

Areas I'd most want checked:

1. **Is the reference map complete?** It counts managed specs, master-YAML entries and base-template slots. Is there a fourth thing in this repo that can legitimately point at a file in `custom_templates/` — something in `template_utils`, the pptx registry's own resolution, a default-template fallback — that I've missed? A miss here means offering to delete a file in use.

2. **The delete guard.** `ctx.orphan(filename)` re-scans and requires an exact match against a name from `iterdir()`. Is that actually sufficient against traversal and symlinks, or is there an input that makes `f.name == filename` true for a file outside the directory? Note there's no `validate_asset_filename()` call — it's kind-specific and this page lists any extension.

3. **Route ordering.** `/files` and `/files/{filename}/delete` are registered before `/{kind}/{name}/…`. Does anything break if someone creates a template kind or a template *named* `files`? And is the ordering guaranteed by registration or am I relying on something incidental?

4. **`scan()` on a hostile directory.** Broken symlinks, a file removed between `iterdir()` and `stat()`, a name with a newline or quote in it reaching the HTML and the confirmation page. Is the unescaped filename in `Code(f.name)` and in the URL safe?

5. **The second-order hazard.** I reproduced it only via a direct POST with an empty `asset_filename`, not through the UI (the draft step overwrites the stale file). Do you agree that's the full reachable surface, and is leaving it to a follow-up the right call for this PR?

---

dvejsada:
Thanks — the Q1 trace is the answer I actually wanted, and it's more thorough than my own: I hadn't checked `xlsx_tools/styles.py` or `_legacy_specs()` by name, and it's worth knowing that every convention-based resolver in the repo lands on one of the five slots rather than that being a coincidence I got lucky with.

**On Q4's URL-construction gap — already fixed, and we landed on the same fix.** The review read `2cc4cf9`; I'd pushed `2438c70` about two minutes before it posted, using exactly `quote(f.name, safe="")` in the two places that build the link. So that one's closed.

One correction to its severity, though, because it matters for whether it was worth fixing: the review says "worst case: a 404 on click, not a bypass". It's slightly worse than that — a `#` makes the browser treat the tail as a fragment, so the request goes to `/admin/files/hash` and the button appears to do *nothing at all*, no error. And the trigger isn't only a hand-placed file: `validate_asset_filename` deliberately allows spaces and non-ASCII, so `Brand Deck.pptx` comes in through an ordinary upload. Agreed it's never a bypass — deletion re-validates against the scan either way.

**Took the Q2 symlink suggestion.** Verified it before writing it down rather than taking it on faith:

- a symlink to a real file *is* listed (`is_file()` follows) and is deletable when unreferenced;
- `unlink()` removes the directory entry, so the target survives — confirmed with a file outside the directory;
- a *dangling* symlink is not listed at all, so it's also not removable here. That's a wart rather than a hazard, and it's now pinned so it stays a decision rather than a surprise.

`db148bd` adds the comment by the `unlink()` call and two tests for those.

**Q3 and Q5** — agreed on both, including that a `save_spec()` fix belongs next to its existing "asset missing" check rather than in this view. I'll raise it as its own issue rather than let it live only in this thread.

`ruff` clean, **2247 passed** (29 new).

---

PR #185, opened 2026-09-20
admin: clone a template

### LINKED ISSUE #164: Admin UI: clone a template
The commonest real task the UI does not support: "same as the formal letter, but for the Prague office". Today that means creating a new template from scratch, re-uploading the document, and re-typing every argument name, type, default and description by hand — where a dozen arguments each with an AI-facing description is normal.

Add a Clone action on the template row and the edit page: prompt for a new name, copy the spec, copy the asset to `<newname><ext>`, save and register. Everything needed exists — `get_spec`, `write_asset`, `save_spec`, `validate_name` — so this is a route plus a small form.

Points to decide in review:

- The asset is copied, not shared. Sharing one file between two specs would make "Replace document" on one silently change the other, which is worse than the duplication.
- For pptx, `default: true` must **not** be carried over — two defaults is an unresolvable state. Drop the key on clone.
- Land the new template on its edit page rather than the index, since the description almost always needs changing immediately and the whole point is to keep going.

dvejsada:
Closes #164.

"Same as the formal letter, but for the Prague office" was the commonest real task the UI could not do. It meant creating a template from scratch, re-uploading the document, and re-typing every argument name, type, default and description by hand — where a dozen arguments each with an AI-facing description is normal.

Clone sits on the template row and on the edit page. Name the copy; everything else comes across. You land on the copy's **edit page**, not the index, because the description almost always needs changing immediately and the point is to keep going.

## The decision that matters: copied, not shared

The issue called this out and I agree with it. Two specs pointing at one file would make "Replace document" on either one silently change the other, and nothing in the UI would report it. The duplication is the cheaper mistake.

So there's a test that actually *does* it — clone, replace the clone's document, then assert the original's wording is still in the original's file. Asserting on the two `docx_path` values would pass just as happily with a shared file if the paths merely looked different.

## Where the logic lives

`store.clone_spec()` owns the copy, the collision check, the name validation and keeping the source's extension (a clone of a `.potx` stays a `.potx`). That's deliberate after #182's review, where the rename path had its safe implementation in the store and a second, untested copy inline in the route — breaking `clone_spec()` fails the route-level tests here, not just a direct one.

Keys a clone must not inherit are declared **per kind** as `clone_drops` in `store._KIND_META`, not branched on in the route:

- **PowerPoint drops `default`.** Two defaults is a state the registry resolves silently by picking one. There's a test that loads the registry after cloning and asserts exactly one default survives — the behaviour, not just the absent key.

## A judgement call worth flagging

**A clone of a disabled template is itself disabled.** The issue says "save and register", which would argue the other way. I went with carrying the flag: `AdminContext.sync()` decides it, the same call that decides it on save, and landing a live tool from a template someone had deliberately taken out of service — with a description that hasn't been edited yet — seemed like the wrong default. The flash says so when it happens. Easy to flip if you disagree.

## Verification

`ruff` clean, **2264 passed** (17 new). Each of these mutations fails the suite:

| Mutation | Result |
|---|---|
| Share the original's asset instead of copying it | 3 failed |
| Carry the pptx `default` flag to the clone | 2 failed |
| Drop the collision check | 1 failed |
| Register unconditionally instead of syncing | 1 failed |
| Land on the index instead of the clone's edit page | 1 failed |
| Don't preserve the source extension | 1 failed |

One process note: the collision mutation first appeared to *survive* — 17 passed. It hadn't been applied. The patch asserted a unique match and the same error string exists in `rename_spec()` too, so the edit silently no-opped and I nearly recorded a false pass. Redone by line, it fails as it should.

Also tightened one of my own weak assertions before pushing: the missing-source-file test originally asserted `"not found" in text or "Asset" in text`, which an `or` makes nearly unfailable. It now asserts the exact message naming the missing file.

## Docs

`admin-ui.md` describes the action. `development/admin-ui.md` gains the copy-not-share and disabled-clone invariants, and the add-a-kind checklist now names `clone_drops` so a new kind declares its own.

---


dvejsada:
@claude review

Areas I'd most want checked:

1. **`clone_spec()` failure modes.** It reads the source asset, then `save_spec()` writes the new asset and the new spec. If the write half fails, what is left on disk — a spec with no file, a file with no spec, or nothing? Is that the right failure, and does the route's `except (TemplateStoreError, OSError)` actually cover it?

2. **The disabled-clone judgement.** I carry `enabled: false` onto the copy, so cloning a disabled template gives a disabled clone. The issue says "save and register", which argues the other way. Which is right, and is there a case where carrying the flag surprises someone badly?

3. **`clone_drops` as per-kind metadata.** PowerPoint drops `default`. Is there anything else in any kind's spec that two templates cannot both hold — something in `layouts`, `style_mapping`, `annotations`, or a key the dynamic-tool loaders treat as unique — that I should be dropping too?

4. **Extension handling.** `f"{new_name}{Path(source_asset).suffix or meta['asset_ext']}"`. Does that hold for a source whose recorded path has no suffix, an unexpected suffix, or a suffix outside the kind's `asset_exts` — and does `validate_asset_filename` still gate it?

5. **The edit page I render after cloning.** I pass `ctx.analyze_asset(kind, copy)` and a flash, rather than redirecting. Does rendering the edit page directly (not a redirect) leave anything inconsistent — CSRF, the `original_name` hidden field, a double-submit on refresh?

---

dvejsada:
The finding is real and it was worse than the clone path. Fixed in `96f0f4d`.

**Root cause, as you traced it:** `sync()` returned `unregister()`'s own bool, which is `True` only when a live tool was actually removed. For a template that's off on purpose there is nothing to remove, so it returned `False`, and callers read that as "could not register".

**You flagged the `save()` conflation as pre-existing and lower priority. I checked, and it's worse than the clone message:**

```
SAVING A DISABLED TEMPLATE -> "Saved with a warning"
                              "Saved, but the tool 'src' could not be registered (check the logs)."
```

That's a bug I shipped in #165 — the disabled-save path is exactly what that PR added, so it went out with a page telling the admin to go read logs about a registration that was never going to happen. Cloning was the second symptom, not the first. Both are fixed together because it's one defect.

`sync()` now answers "does the server match the spec?", so taking a template off is success. Saving has three outcomes rather than two — live, deliberately off, failed — and `saved_page()` picks between `save_ok`, a new per-kind `save_disabled` and `save_warn` from the enabled state instead of inferring it from one boolean:

| | before | after |
|---|---|---|
| Save, enabled | ✓ Saved — now live | unchanged |
| Save, disabled | ⚠ **could not be registered (check the logs)** | ✓ Saved. Stays disabled — enable it when you're ready |
| Clone of disabled | …**(not live yet — save to register it)** | …It is disabled, like the template it was copied from |

**Worth reporting: two of my mutations for this survived the first run, both my own test weakness rather than surviving code.**

1. `assert "disabled" in r.text` is unfailable — the inlined theme CSS carries `--ctl-disabled-bg` and `input[disabled]`, so every admin page contains the word. It now asserts the actual sentence.
2. Once the messages stopped keying off `sync()`'s return value, nothing observed it, so the `return True` was untestable through the UI. Rather than drop it — leaving `sync()` reporting success as failure for the next caller — its contract is now pinned directly.

Both fail the suite now.

**On your other points:** agreed on Q1's TOCTOU being pre-existing and low-severity for a single-password UI; Q3's confirmation that `load_specs()` never collapses two `True` defaults is the detail that makes `clone_drops` worth having rather than cosmetic; Q4 and Q5 match what I found testing them myself.

`ruff` clean, **2270 passed**.

---

PR #186, opened 2026-09-20
admin: inspect and adopt hand-written master-YAML templates

### LINKED ISSUE #167: Admin UI: master-YAML templates are listed read-only with no way to inspect or adopt them

dvejsada:
Closes #167.

A template defined in `config/<kind>_templates.yaml` was a bare row: the name, a dash, "● Live", and the badge `from master YAML — read-only`. That was the entire affordance — no description, no arguments, no style mapping, no sight of which file it used — even though `gather_specs()` had returned the whole spec to build that row. For anyone whose templates predate the admin UI, that was most of their templates, permanently second-class.

## Inspect

The name links to a read-only page: description, arguments **with their descriptions**, style mapping, the source file and which directory it actually resolves from, plus the same analysis the edit page runs. No form, no save.

```
legacy_letter
  Status       ● Live
  Description  Predates the admin UI
  Source file  default_docx_template.docx — default_templates/
  Arguments    1
```

## Adopt

Copies the entry into the managed `*.d` layer, after which it edits like any other template. `gather_specs()`' merge order does the work, so **the master file is never rewritten** — the test asserts it stays byte-identical, not merely that it still parses.

Where the document ships in `default_templates/`, it's copied into `custom_templates/`. Otherwise you'd have a template you can edit but whose document you can never replace — the confusing half-state the issue warned about.

## One change beyond the issue

The master rows now come from `_master_specs()` rather than from the live tool names. Keyed off what's registered, a master template that is **disabled** — or that failed to load — vanished from the page entirely: no way to inspect it, no way to adopt it. That's the worst case to hide, and since #165 gave `enabled: false` to master entries too, it's newly reachable. Flagging it as scope I added rather than burying it.

## Verification

`ruff` clean, **2287 passed** (17 new). Mutations caught: dropping the already-managed guard; never copying a shipped file into the writable directory; listing from live names again; and omitting each of the four detail cards.

**Two mutations survived the first run. Both were my tests, and they failed differently:**

1. **Vacuous assertions.** Every detail-page check passed via the "as written" YAML block, which serialises the *whole* spec — so I could delete the Arguments card, the Style mapping card or the analysis and still find the text on the page. All four now assert above that block. This is the same class of mistake as the `"disabled" in r.text` one in #185, and it's now bitten twice, so it's worth naming: **asserting a value appears somewhere on a page proves nothing when the page also dumps the source data.**

2. **A genuine equivalent mutant.** "Adopt does not recopy a file already in `custom_templates/`" cannot be tested: `_candidate_dirs()` searches custom first, so a recopy resolves to the same file and rewrites it byte-for-byte. The guard is an I/O optimisation, not a correctness property. I kept it and rewrote the test to state plainly what it does and does not cover, rather than leave a test whose name claims more than it checks.

## Docs

`admin-ui.md` describes both actions. `development/admin-ui.md` records why adopting is additive and why the rows are listed from the YAML rather than from the registry.

---


dvejsada:
@claude review

Areas I'd most want checked:

1. **Does adopting actually leave the master YAML alone in every case?** The test asserts byte-identity for one adopt. Is there any path — `save_spec`, `sync`, the registry reload, a later edit of the adopted copy — that writes back to `config/<kind>_templates.yaml`?

2. **The adopted copy winning.** I rely on `gather_specs()` preferring the `.d` entry by name. Does that hold for the PowerPoint registry too, and for a master entry that is *later* renamed or disabled in the YAML by hand while an adopted copy exists?

3. **Asset copying on adopt.** When the document lives in `default_templates/`, I copy it into `custom_templates/` under the same filename. That filename is shared with the base-template slots (e.g. `default_docx_template.docx`). Does adopting a master entry that names a base-template file create a collision with the Base templates page or the Source files reference map from #166?

4. **Listing from `_master_specs()` instead of live names.** I changed this so disabled master templates stay visible. Does it introduce a row that should not be there — a malformed entry, an entry with no name, or one whose file is missing entirely?

5. **The read-only page's honesty.** It renders `spec_yaml_block(spec)`, whose text says "This is what the UI wrote into the template directory" — which is false here, since the UI wrote nothing. Is that misleading enough to need its own wording for this page?

---

dvejsada:
Self-answering my Q3 before the review gets to it, because the answer was **yes, and it was dangerous**. Fixed in `dd37717`.

Adopting a master entry that points at `default_docx_template.docx` copied that file into `custom_templates/`. `template_utils` searches the custom directory first, so the base Word template silently moved:

```
base docx template BEFORE:        …/default_templates/default_docx_template.docx
base docx template AFTER adopt:   …/custom/default_docx_template.docx
```

Byte-identical, so harmless on its own. But the adopted template is now editable, and using **Replace document** on it rewrote that file:

```
base docx template AFTER replace: …/custom/default_docx_template.docx
  -> content now: "WRONG BASE STYLING"
```

One template's document replacement silently restyling **every** Word document the server generates. Reproduced end to end before touching anything.

`base_templates.RESERVED_FILENAMES` now names every such file — both the `custom_` and the `default_` spelling of all five slots, since either one shadows — and adopting an entry that points at one gives the template a private copy under its own name instead. Two tests cover the consequence, not just the mechanism: the base template still resolves to the shipped file after adopting, and replacing the adopted template's document leaves it untouched.

I also had to replace one of my own tests: "copies a shipped file into the writable directory" used `default_docx_template.docx`, which is exactly the case now forbidden. It moved to `broadcast_email_style_1.html` — also shipped in `default_templates/`, but not a base-template name — so the ordinary path stays covered.

Mutations: allowing a reserved filename through, and reserving only the `custom_` spellings rather than both. Both fail the suite.

**Q1, Q2, Q4 and Q5 are still genuinely open** — particularly Q2 (whether the `.d` entry keeps winning if the master entry is later renamed or disabled by hand) and Q5 (the YAML block's caption claims "this is what the UI wrote into the template directory", which is false on a page showing a file the UI never wrote).

`ruff` clean, **2289 passed**.

---

dvejsada:
Acted on this in `675f0bd`. Note the review read `4cfc599`, before `dd37717`.

**Q3 — we found the same bug independently.** You reasoned to it from `SLOTS`; I reproduced it end to end from the other direction. Same conclusion, already fixed in `dd37717` with `RESERVED_FILENAMES` covering both the `custom_` and `default_` spelling of all five slots. Worth noting your example (`docx_path: custom_docx_template.docx`) and mine (`default_docx_template.docx`) are both live — which is exactly why reserving only one spelling isn't enough, and there's a mutation for that.

**Q2 — took it. Reproduced both consequences first:**

```
after adopt   : ['legacy_letter']
after rename  : ['legacy_letter_v2', 'legacy_letter']
after disable : ['legacy_letter']
```

So both your readings are right: renaming leaves two templates, disabling is a silent no-op.

Where I'd push back slightly is on framing the rename as a gap in *this* PR. It's how a `.d` entry has always shadowed a master entry of the same name — creating a managed template with a master template's name has done exactly this since the merge layer existed. What adoption changes is that it's now one click, which is what makes it worth stating rather than fixing here: following a rename means recording provenance on the copy, which is a larger feature than adopting is.

So instead: the adopt card now says plainly that the copy stops following the entry ("matched by name… later edits to this entry in the YAML, including turning it off, no longer have any effect"), both behaviours are pinned by tests so a change to either is deliberate, and `development/admin-ui.md` records why provenance tracking was left out. If you'd rather have the provenance, say so and I'll file it.

**Q5 — you didn't flag it, but I asked and it was real.** The YAML block said *"this is what the UI wrote into the template directory"* on the one page whose entire point is that the file belongs to the admin. Now conditioned, with a test in each direction so the wording can't flip.

**Q1 and Q4** — thanks for tracing every write path rather than spot-checking; "the byte-identity test is testing a real invariant, not a coincidence" is the distinction I wanted.

`ruff` clean, **2293 passed**.

---

Issue #149, opened 2026-09-19
Admin UI: template lifecycle operations beyond create / edit / delete


dvejsada:
The UI supports exactly four verbs: create, edit, replace-document, delete. Everything an admin does with a template *after* it works is unsupported, so the workflows that actually recur — spin a variant off an existing letter, take a template out of service for a week, tidy up the files a delete left behind, bring a hand-written YAML template under UI management, prove a template against real text rather than `[recipient_name]` — all mean editing YAML and copying files by hand.

Delete is the only off switch, which is the sharp edge: taking a tool away from the AI for a day means destroying its configuration.

Sub-issues below.

PR #187, opened 2026-09-20
admin: preview a template with your own values

### LINKED ISSUE #168: Admin UI: preview with your own values, not just auto-generated samples
Preview fills every argument automatically. `admin/preview.py:26` `sample_values()` substitutes the declared default when non-empty, otherwise a bracketed name like `[recipient_name]`, and forces every boolean to `True` so conditional blocks are visible.

That proves the placeholders are wired up. It does not prove the template works, because the one thing these templates are for — Markdown inside a placeholder — is never exercised. `docs/templates.md` promises full Markdown in placeholder values, with block-level Markdown rendering only for body placeholders and inline-only in table cells, headers and footers. `[body]` is a plain string, so the preview renders a plain string, and the admin ships a template having never seen a list, a heading or a bold run come out of it. The distinction between body and cell placeholders — the one that actually bites — is invisible.

Forcing all booleans to `True` hides the other half: no preview ever shows the document with a conditional block **off**, so an unbalanced or mis-scoped `{{#if}}` looks fine until a real call turns it off.

Add a preview form: one input per declared argument, pre-filled with the current sample value so it stays one click for anyone who does not care, checkboxes for booleans so conditionals can be toggled. Submit to the existing preview route with the values in the form instead of generating them.

Applies to docx and email. PowerPoint previews a fixed six-slide deck (`SAMPLE_DECK`, `preview.py:88`) and needs nothing here — a presentation template takes no arguments.

dvejsada:
Closes #168 — and with it **#149**, whose five sub-issues are now all done.

Preview filled every argument automatically: a bracketed `[name]` for each string, the declared default for each flag. That proves the placeholders are wired up. It does not prove the template *works*, because the one thing these templates exist for — Markdown inside a placeholder — was never exercised. `[body]` is a plain string, so the preview rendered a plain string, and an admin could ship a template having never seen a list, a heading or a bold run come out of it.

**Preview with my values…** opens a form with one control per argument, pre-filled with the same samples so it stays one click for anyone who doesn't care. Text values are **textareas** — a single-line input cannot hold the list this feature exists to try. Conditionals are checkboxes, so a block can be previewed off as well as on.

Verified by rendering the document, not by inspecting the form:

```
value_body = "# A Heading\n\n- first\n- second\n\nSome **bold** text."

   [Heading 1     ] 'A Heading'
   [List Bullet   ] 'first'
   [List Bullet   ] 'second'
   [Normal        ] 'Some bold text.'  BOLD
conditional block shown (unchecked): False
```

## Three things a sample could never show, not two

I expected Markdown-through-a-placeholder and conditionals-off. `docs/templates.md` also promises block-level Markdown in a *body* placeholder but inline-only inside a **table cell** — and with a bracketed sample neither treatment is exercised, so the difference was invisible. Typing the same list into both makes it plain:

```
BODY paragraphs:
   [List Bullet ] 'one'
   [List Bullet ] 'two'
CELL paragraphs:
   [Normal      ] '- one\n- two'
```

## A correction to the issue's premise

The issue says samples force *every* boolean to True, so a block can never be previewed off. That's only half right, and I found it because a test I'd written from the issue's description failed:

| flag | sample | so it could never be seen… |
|---|---|---|
| undeclared conditional | `True` | off |
| declared, **required** | `True` | off |
| declared, **optional** | `False` | **on** |

`sample_values()` uses the declared default and only falls back to `True` when there is none — and `build_spec()` writes `default: false` for an optional flag with a blank default. So an optional conditional had the opposite problem to the one reported. All three are now pinned, and `sample_values()`' docstring — which claimed flags default to True — says what it actually does.

## The bug the review caught

An unticked checkbox sends **nothing**. A template whose arguments are all booleans therefore submits no `value_` key when every box is off — and absence of `value_` keys meant "generate samples". So on the simplest template that has a conditional, unticking it came back with the block rendered **on**:

```
all-boolean template, box UNTICKED -> block shown: True
EXPECTED: False (the admin unticked it)
```

The one thing this feature exists to make possible, silently doing the opposite. The form now carries `preview.VALUES_MARKER` so a submission is recognised whatever its controls happen to send; the prefix check stays so a hand-built POST still counts.

## How the values reach the renderer

The form is reached *from the edit form*, so it must preview what's on screen, unsaved edits included. Re-emitting every field `build_spec()` reads would duplicate its contract and drift from it, so the rebuilt spec rides along as JSON in one hidden field; `carried_spec()` ignores anything that isn't a named mapping.

That spec is client-supplied, so its asset path is attacker-shaped — but `asset_exists()` resolves through `asset_path()` → `validate_asset_filename()`, which refuses anything but a bare filename. Five traversal shapes all return 400 with the target file's bytes nowhere in the response; there's a test. An apostrophe in a description also round-trips correctly, because "the client's letter" is an ordinary description.

## Verification

`ruff` clean, **2328 passed** (35 new). Mutations caught:

| Mutation | Result |
|---|---|
| Submitted values are ignored | 2 failed |
| A boolean keeps its sample instead of reading the checkbox | 1 failed |
| The carried spec is ignored | 2 failed |
| A string gets a single-line input | 1 failed |
| An undeclared conditional gets no control | 1 failed |
| `has_submitted_values` ignores the marker | 2 failed |
| The form stops emitting the marker | 1 failed |

## Docs

`admin-ui.md` describes the form and why the textareas matter. `development/admin-ui.md` records how the spec travels between the two POSTs, why the marker is needed, and what a sampled boolean actually does.

---


dvejsada:
@claude review

Areas I'd most want checked:

1. **`carried_spec()` is client-supplied JSON that drives a render.** It only reaches `render_docx_preview` / `render_email_preview` / `build_style_map`, behind the admin password, and `build_spec()`'s output is equally client-supplied. Is there a key in a spec — `style_mapping`, `annotations`, the path key — where a crafted value does something worse than produce a silly document? Specifically: can the path key escape `custom_templates/`?

2. **Value coercion.** `values_from_form()` falls back to the sample for anything missing or unparseable, and honours an empty string as empty. Is there an input where the fallback is wrong — a declared `int` given `"0"`, a string arg whose sample is itself a bracketed name, a value for an argument that isn't declared at all?

3. **The boolean rule.** A checkbox's absence means False, so every unticked flag is False rather than "unset". Is there a case where a template distinguishes those, or where reading the checkbox for an arg that is a *declared bool but not a conditional* changes the rendered document in a way I haven't considered?

4. **Email preview.** I wired values through the shared path, but the email renderer is pystache rather than the docx pipeline. Does an arbitrary string value behave there — Markdown, HTML, a `{{` sequence in the value itself?

5. **Did I miss a third thing samples can't show?** I found Markdown-through-a-placeholder and conditionals-off. `docs/templates.md` also distinguishes body placeholders (block-level Markdown) from table cells, headers and footers (inline only). Is that distinction now testable through this form, or still invisible?

---

dvejsada:
Self-answering Q1 and Q5 in `66b8801` while the review runs.

**Q1 — the carried spec cannot read outside the uploads directory.** `asset_exists()` resolves through `asset_path()`, which runs `validate_asset_filename()` and refuses anything but a bare filename, so a crafted path never reaches `read_asset()`. Five shapes, all 400, with the target file's bytes nowhere in the response:

```
../secret.docx             -> 400
../../secret.docx          -> 400
/etc/passwd                -> 400
../config/secret.yaml      -> 400
sub/../../secret.docx      -> 400
```

Now pinned by a test, since this is the one genuinely security-relevant surface the feature adds.

**Q5 — yes, and it's the third thing.** `docs/templates.md` promises block-level Markdown in a body placeholder but inline-only inside a table cell. With a bracketed `[body]` sample *neither* treatment is exercised, so the difference was invisible. Typing the same list into both makes it plain:

```
BODY paragraphs:
   [List Bullet ] 'one'
   [List Bullet ] 'two'
CELL paragraphs:
   [Normal      ] '- one\n- two'
```

So the form surfaces three things samples couldn't, not two. The test asserts the cell was rendered *at all* before asserting what it lacks — "no bullets" is equally true of a cell that never rendered, and that's the shape of vacuous assertion that has already bitten me twice in this series.

**Q2, Q3 and Q4 are still open** — particularly Q4, since the email path is pystache rather than the docx pipeline and I've only exercised it through the shared code path, not with a `{{` sequence inside a submitted value.

`ruff` clean, **2326 passed**.

---

dvejsada:
Your checkbox-absence finding is a real bug, and it's in the narrowest case — the one this feature exists for. Fixed in `2f40f7c`.

An unticked checkbox sends nothing at all, so a template whose arguments are **all booleans** submits no `value_` key when every box is off. Absence of `value_` keys meant "generate samples". Reproduced on a template whose sole argument is a conditional:

```
all-boolean template, box UNTICKED -> block shown: True
EXPECTED: False (the admin unticked it)
```

So the one thing #168 set out to make possible — previewing a block *off* — silently did the opposite on the simplest template that has one. The form now carries `preview.VALUES_MARKER`, so a submission is recognised whatever its controls happen to send; the `value_` prefix check stays so a hand-built POST still counts. Two mutations cover it.

I'd flag that this is the most valuable finding in the series, and it came from a question about a *stringification* detail I'd asked badly — Q3 was aimed at how booleans reach the renderer, and the actual bug was one layer earlier, in whether the submission was recognised at all. Worth noting for how I phrase these.

**On the sandbox note:** CI ran the suite — green on every head, currently **2328 passing** on `2f40f7c`. I also ran `ruff` and the full suite locally before each push, and reproduced every finding by execution rather than reading. So the review's caveat is covered.

**Q1, Q2, Q4, Q5** — all match what I found testing them, with two details I hadn't checked and am glad to have: that `"0"` survives because the guard is `raw is None` rather than a truthiness test, and that pystache substitutes plain data so a `{{` inside a submitted value is inserted literally rather than re-parsed. The point that `carried_spec()` changes *how* an admin-supplied dict arrives but not its trust level is the right framing — the pre-PR route already built the spec from raw form fields.

Also confirming your Q5 answer independently: `replace_placeholders_in_document` passes `doc=None` for cells and headers, and the rendered output shows it — body becomes two `List Bullet` paragraphs, the cell keeps `'- one\n- two'` as one `Normal`.

---

Issue #150, opened 2026-09-19
Admin UI: manage the template kinds it cannot touch at all today


dvejsada:
The UI manages dynamic templates — the ones that become tools, plus registered PowerPoint designs. It manages none of the **static** templates, which are the most commonly customised files in the product:

| File | What it controls |
|---|---|
| `custom_docx_template.docx` | the look of every generated Word document |
| `custom_email_template.html` | the wrapper on every generated email |
| `custom_pptx_template_4_3.pptx` / `_16_9.pptx` | the default deck design |
| `custom_xlsx_template.xlsx` | the named cell styles `style:<Name>` directives resolve against |

Changing any of them means shelling into the volume and copying a file — exactly the friction `docs/admin-ui.md` says the UI exists to remove. Excel has no admin presence whatsoever.

Two smaller gaps in the same category: the pptx analysis report omits `content_area`, which `docs/templates.md` calls a key diagnostic, and the upload limit will reject the brand decks this feature is for.

Sub-issues below.

PR #188, opened 2026-09-20
admin: report the pptx content area

### LINKED ISSUE #171: Admin UI: the pptx analysis report omits content_area
`_pptx_analysis_report` (`admin/app.py:593`) reports slide size and aspect, every layout with its placeholders and detected role, footer coverage, and the theme fonts and colours. It does not report `content_area`.

`docs/templates.md` treats that as a first-class diagnostic, and `list_presentation_templates` returns it under `include_layouts`:

> The `content_area` is read from the template's own content layout. Slides the tool positions itself stay inside it — a `blank` or title-only layout keeps the template's logo, rules and footer, so drawing from the top-left corner would land on them.

So it is the rectangle that decides whether generated content lands on the template's own furniture. If it is read from the wrong layout, or comes out implausible on a template whose content layout is unusual, every self-positioned slide is subtly wrong — and the admin UI, which exists to make exactly this kind of mis-detection visible before a deck is generated, does not show it.

Add it to the facts block: the layout it was read from and the four percentages, formatted as `docs/templates.md` shows them. Worth a warning when the area is implausible (very small, or nearly the whole slide, which usually means no content layout was found and a default was assumed) — that is the case an admin can actually act on, by setting a layout override.

dvejsada:
Closes #171 — and with it **#150**, whose four sub-issues are now all done.

The pptx analysis card reported slide size, every layout with its placeholders and detected role, footer coverage and the theme — but not `content_area`, which `docs/templates.md` treats as a first-class diagnostic and `list_presentation_templates` already returns.

It's the rectangle generated content is positioned against. A `blank` or title-only layout is rarely an empty canvas: templates keep a logo, a header rule and a footer band on it. So if this rectangle is read from the wrong layout — or missing — every self-positioned slide lands on that furniture. The page that exists to make exactly this kind of mis-detection visible *before* a deck is built was the one place it couldn't be seen.

```
Slide size     13.33 × 7.50 in (16:9)
Content area   x 6.9%  y 26.6%  w 86.2%  h 63.4%  — read from 'Nadpis a obsah'
```

## One formatter, not two

`_content_area_summary()` is now public rather than copied into the admin layer. That matters more than it sounds: the page and `list_presentation_templates` describe the same rectangle in the same units, and if they ever disagreed there'd be nothing to say which was right. A test asserts they match on the shipped template — and that value also matches the worked example printed in `docs/templates.md` **verbatim**, which is a second, independent check that the reading hasn't drifted.

## Three warnings, each actionable

| Case | Why it matters |
|---|---|
| No layout declares one | The builder falls back to a fixed 1.5in band that knows nothing about this template's decoration |
| Covers ~the whole slide | Positioned content stays clear of nothing |
| Implausibly small | Content will be cramped |

Each names the layout it was read from, so it can be checked directly.

Read with **no layout overrides applied**, like the role map beside it — the overrides are on the form next to the report, so the card shows what the template says on its own. Documented rather than left to be inferred.

## Verification

`ruff` clean, **2339 passed** (11 new). Mutations caught:

| Mutation | Result |
|---|---|
| Never read the content area | 3 failed |
| Omit the row from the report | 2 failed |
| Don't warn on a whole-slide area | 1 failed |
| Don't warn on a tiny area | 1 failed |
| Divide by zero on a malformed slide size | 1 failed |
| **Small-area threshold far too eager** | 3 failed |

That last one is deliberate: a warning that fires on ordinary templates is worse than no warning, so the thresholds are mutated in the crying-wolf direction too, against three shapes real templates actually have.

## Docs

`admin-ui.md` describes what the card now shows. `development/tools/powerpoint.md` records why the formatter is public. `development/admin-ui.md` records the no-overrides reading and what the warnings cover.

---


dvejsada:
@claude review

Areas I'd most want checked:

1. **The no-overrides reading.** `_read_content_area()` builds `LayoutResolver(presentation, {})`, so a template whose spec sets a `layouts:` override for a content role may resolve a *different* rectangle at build time than the card shows. I judged that acceptable because the override is on the form beside the report — is it, or is the card actively misleading in that case?

2. **The thresholds.** `>= 0.95` both dimensions for "whole slide", `< 0.25` either dimension for "too small". Are those defensible against real templates, or will one of them fire on an ordinary design — a wide banner content area, say, or a narrow sidebar layout that is legitimately the content reference?

3. **Making `content_area_summary()` public.** It now has two callers with different lifecycles. Is there anything in it that assumed a single internal caller — the `round(…, 1)` formatting, the `rect.source` being a layout name, the `None` returns?

4. **`_read_content_area()` is called inside the `TemporaryDirectory` block** in `analyze_pptx`, after `_read_theme`. Is that the right place — does anything it touches need the staged file to still exist, and would it be safer outside?

5. **The warning I added for "no content area at all".** It fires whenever no layout has a body placeholder. Is that genuinely rare enough to be a useful signal, or common enough on real-world decks that it becomes noise on the analysis card?

---

dvejsada:
Self-answering Q1 and Q4 in `075f656`, and Q1's answer is better than the assumption I'd built on.

**Q1 — the card is exact, not approximate.** I'd documented the no-overrides reading as a deliberate simplification. It isn't one: `LayoutResolver` applies a configured `layouts:` mapping in `provides()` and `resolve()`, but `content_area()` reads `_by_role`, which is filled from `classify_layout()` alone. `self._configured` never reaches it:

```
resolver._by_role['content'] with {"content": "Dva obsahy"} -> Nadpis a obsah
```

So an override doesn't move the rectangle at build time either, and the card shows exactly what the builder uses. Now pinned by a test that picks a layout whose rectangle *genuinely* differs from the resolved one — the shipped template has seven distinct content rects across eleven layouts, so the assertion can't pass just because two happen to match.

**That turned up a real product inconsistency**, which I've filed as #189 rather than fixed here: an admin who overrides the content role because detection picked the wrong layout has fixed where `content` slides are built, but *not* the rectangle `blank`/KPI/timeline slides position themselves against — which is the problem the override was set to solve. Changing it alters generated geometry for existing templates, so it wants a deliberate decision, not a drive-by in a reporting PR. The test here names the requirement: if overrides start affecting the reading, the card has to follow.

**Q4 — placement is fine.** Nothing in the reading needs the staged upload to still exist; python-pptx loads the package into memory at open time, and the rectangle reads correctly after the `TemporaryDirectory` is gone. Inside or outside the block is equivalent, so I've left it where it is, next to the other reads.

**Q2, Q3 and Q5 remain genuinely open** — particularly Q2, since my thresholds are judgement calls I can only test against the one template that ships here, and Q5, where I have no corpus to say whether "no content area at all" is rare enough to be a signal.

`ruff` clean, **2340 passed**.

---

dvejsada:
Q2's residual gap is real and now fixed in `af162f3`. Note the review read `38a0616`, two commits back.

**Q2 — you're right, and it false-positives on two shapes, not one.** `w < 0.25 or h < 0.25` calls a narrow full-height rail cramped; checking properly, a wide shallow banner trips it too:

```
shape                     w      h    area  or-rule  area-rule
shipped 16:9           0.86   0.63   0.547     ok       ok
narrow sidebar         0.20   1.00   0.200   WARN       ok
wide banner            1.00   0.18   0.180   WARN       ok
half-width column      0.48   0.64   0.307     ok       ok
genuinely tiny         0.10   0.10   0.010   WARN     WARN
```

Both are designs a template may legitimately use as its content reference. What actually makes content cramped is how little room there is *in total*, so the rule is now `w * h < 0.08`. The cry-wolf matrix gains the sidebar, the banner and a two-column half. Mutating back to the `or` rule now fails on the first two; dropping the rule fails on the tiny one.

I'd push back gently on "low severity because the warning is advisory". On a page whose whole job is to make mis-detection visible, a warning that fires on ordinary templates is worse than no warning — it teaches the admin to skim past the card, which is where the *real* signals are. That's why the thresholds get mutated in the crying-wolf direction too.

**Q1 — same conclusion, and the wording point was already half-fixed.** I'd reached it independently in `075f656` (after you started reading): `_by_role` is `classify_layout()` alone, so passing overrides picks the same rectangle. But you're right that the doc still *led* with "the overrides are on the form next to the report", which implies applying them would differ. Rewritten to lead with the actual reason. I've also filed **#189** for the underlying inconsistency — an admin overriding the content role fixes where `content` slides are built but not the rectangle `blank`/KPI/timeline position against, which is the problem the override was set to solve.

**Q3 — taking the analysis, leaving the redundancy.** `resolver.content_area()` twice is deliberate: one call wants the `Rect` for the thresholds, the other wants the formatted summary, and `content_area()` caches. Collapsing them would mean either returning a tuple from the formatter or duplicating the percentage maths — both worse than one cached call.

**Q4 and Q5** — Q4 I'd already verified (nothing needs the staged file; python-pptx loads the package at open time). Q5 I still can't answer from one template, so the warning stays as written; if it turns out noisy on real decks it's a one-line threshold, not a design change.

`ruff` clean, **2343 passed**.

---

PR #190, opened 2026-09-20
store: do not adopt an asset whose name was only guessed

### LINKED ISSUE #184: save_spec adopts a stale same-named asset when asset_filename is empty
Split out of #166, where it was noted as a second-order hazard. Reproduced and scoped rather than assumed.

`FileTemplateStore.save_spec()` derives the asset filename from the template name when neither an explicit filename nor a `path_key` is given:

```python
filename = asset_filename or spec.get(path_key) or f"{name}{meta['asset_ext']}"
```

and then only raises when the asset is **missing**:

```python
elif not self.asset_exists(kind, filename):
    raise TemplateStoreError(...)
```

So if a file of that name is already sitting in `custom_templates/` — left behind by a template that was deleted with its source kept — the new spec silently adopts it. The admin gets a working template built on a document they never uploaded in this session, with nothing saying so.

### Reachability

Narrower than the original note in #166 suggested:

- **Through the UI: not reachable.** The two-step create flow uploads via `/{kind}/draft`, which writes the file to `<name><ext>` before `/{kind}/save` runs — so the stale file is *overwritten*, not adopted. Verified by reading back the resulting document's text: it is the newly uploaded content.
- **By POSTing `/{kind}/save` directly with an empty `asset_filename`: reachable.** The spec is created and points at the stale file. Verified.

The edit form always carries `asset_filename` as a hidden field, so nothing in the browser flow hits this.

### Why it is still worth closing

It is a latent trap in the store's own contract rather than in one caller: the guard reads "the asset must exist", when what the caller means at that point is "the asset must be one I just wrote". A future caller that builds a spec without going through the draft step inherits the behaviour silently.

### Suggested fix

Beside the existing "asset missing" check in `save_spec()`, not in the view layer — per the review discussion on #183. Either:

- refuse to derive a filename onto an existing file when no asset bytes were supplied, so adoption has to be explicit; or
- keep allowing it but make the caller pass the filename it means, so the derive branch never silently resolves to someone else's leftovers.

### Mitigation already in place

#183's **Source files** page makes the aftermath visible: the reused file shows up correctly attributed to whichever template now points at it. That is after the fact, not prevention.

dvejsada:
Closes #184.

`save_spec()` takes the asset filename from its argument, else the spec's path key, else the template's name. **That last branch is a guess** — and the guard beneath it read "the asset must exist", when what a caller in that branch actually means is "the asset must be the one I just wrote".

So a file left behind by a template deleted with its source kept was adopted by the next template of the same name, silently: a working template built on a document nobody uploaded in this session, with nothing saying a file had been *reused* rather than received.

```
derive onto a leftover -> 'report.docx' already exists, and no document was supplied
                          for 'report'. Upload one, or name the file explicitly if
                          reusing it is what you mean.
  spec created: False
```

## What is and isn't affected

Only the guess. Naming the file explicitly still reuses it — which is how **every** edit of an existing template saves, and how `clone_spec()` and `adopt()` pass their copies. The whole suite passed unchanged after the fix, which is the useful signal here: nothing relied on the adoption.

| Call | Before | After |
|---|---|---|
| Derived name, no bytes, file exists | silently adopts it | **refused** |
| Derived name, no bytes, no file | "does not exist yet" | unchanged |
| Derived name **with** bytes | writes them | unchanged |
| Explicit `asset_filename` | reuses | unchanged |
| Spec carrying its own `path_key` | reuses | unchanged |

The two refusals stay distinguishable — *nothing there* and *something unexpected there* are different problems, and a shared message would send an admin looking for the wrong one. There's a test asserting they differ.

## Reachability is unchanged from the issue

Through the UI the draft step writes the upload before `save` runs, so nothing in the browser flow ever reached this. It was a trap in the store's own contract, waiting for the next caller that builds a spec without going through the draft step — which is exactly why it belongs in `save_spec()` rather than in a route.

## Verification

`ruff` clean, **2350 passed** (7 new). Mutations caught:

| Mutation | Result |
|---|---|
| Remove the guard entirely | 3 failed |
| **Apply it to explicitly named files too** | 39 failed |
| Let the derived name count as "named" (silently disables it) | 3 failed |

The middle one is the one worth having: refusing legitimate reuse would be considerably worse than the bug, so the tests are mutated in that direction too.

## Docs

`development/admin-ui.md` records what the guess is, why the old guard was the wrong question, and that explicit naming still reuses.

---


dvejsada:
@claude review

Areas I'd most want checked:

1. **Is `named` the right test for "guessed"?** It's `asset_filename or spec.get(path_key)`. A spec whose `path_key` is present but *empty* (`""`) falls through to the derived name — is that the behaviour I want, or should an explicit empty string be an error rather than a silent guess?

2. **Did I break a legitimate flow?** The suite passed unchanged, which I read as "nothing relied on adoption". Is there a path — a test helper, a future caller, the pptx registry — where deriving onto an existing file was the *intended* convenience, and the refusal is now just friction?

3. **The two error messages.** One says "does not exist yet", the other "already exists, and no document was supplied". Is the second one's advice actually actionable from where it surfaces — the save route renders it on `save_failed_page`, where "name the file explicitly" may mean nothing to an admin who has no field for that?

4. **`clone_spec()` and `adopt()` both pass `asset_filename` explicitly**, so they're unaffected. But `clone_spec()` reads the source asset and passes bytes *as well*. Is there a case where the two disagree — bytes supplied but the filename points somewhere else?

5. **Anything else in `save_spec()` that asks the wrong question?** The `named` fix is one guard whose condition didn't match its intent. The `validate_asset_filename` call, the `mkdir`, the order of asset-then-spec — do any of those have the same shape of mismatch?

---

PR #191, opened 2026-09-20
admin: edit the global style mapping

### LINKED ISSUE #161: Admin UI: the global style_mapping is read-only

dvejsada:
Closes #161. Last of the admin-UI issues.

> **Three follow-up commits since this was opened** — see the thread. The sections marked ⚠️ below describe the *first* commit and were superseded; everything else stands.

The `style_mapping` that applies to **every** Word document was the one setting the UI could only read — the one that rebrands every document the server produces, and the one that still needed hand-editing YAML on the volume.

Part 1 (each "(use built-in)" option naming what the key actually resolves to) shipped with #171. This is part 2.

## A home for global config in the merge layer

The master YAML is hand-written documentation the tooling never rewrites, so the override goes in the same `*.d` layer the per-template specs use, as a reserved `_global.yaml`. `template_registry.global_config()` is the one merge point for it, as `gather_specs()` is for templates.

**A key there replaces the master's key; it does not merge into it.** This is the decision the whole feature rests on:

| | key-wise merge | replace (chosen) |
|---|---|---|
| Admin clears `heading_1` | falls back to the master's `Brand Title` | built-in `Heading 1` |
| What the page shows | `(use built-in: Heading 1)` | `(use built-in: Heading 1)` |
| | **the page is lying** | the page is right |

A merge would reproduce the exact lie #161 is named after, one layer down. Presence is what counts, so an explicit empty mapping means *nothing overridden* while deleting the file hands the kind back to the master — two different states, and the page offers both (Save with everything cleared, and Revert). `templates:` is never taken from the override; one file able to register tools by a back door is not a power this layer should have.

The filename is reserved in both directions: `read_spec_dir()` skips it so it is never read as a template, and `validate_name()` refuses the name `_global`. Every store operation that names a spec builds its path through `_spec_path` → `validate_name`, so all six doors (`save_spec`, `rename_spec`, `clone_spec`, `set_enabled`, `get_spec`, `delete_spec`) refuse it, and there is a test that says so.

## Three consumers that had to move together

The mapping has three readers and they resolved it by different routes:

| Consumer | Resolved | Picked up an edit |
|---|---|---|
| `markdown_to_word` (static) | once per process, **cached forever** | never, short of a restart |
| a dynamic Word template tool | once, at registration | never |
| the admin preview | per render | immediately |

So a naive "write the YAML" implementation would have produced a page showing the new mapping while every generated document went on using the old one.

- `load_global_style_map()` now caches against the config files' **mtime and size**, never for the process lifetime. (⚠️ The first commit dropped the cache entirely. Measured, that cost **~8 ms** per document against a ~50 ms markdown render — not the "negligible" I claimed. A cache hit is ~0.02 ms.) The fingerprint cannot see two writes inside one filesystem timestamp tick that leave the file the same length, so the *writer* says so: `invalidate_global_style_map()`, called from `resync_docx_style_map()`.
- Saving **re-registers every Word template tool**, because they bake the map in. Without it a save would move the static tool and leave the template tools behind.
- **A template lost to that re-registration is named, as a warning.** `register_docx_template()` removes the existing tool *before* rebuilding it, so a template whose source file has gone missing since startup is not merely stale — it is off the server. Reproduced; the page used to report it as `0 tool(s) re-registered` and nothing else. `resync_docx_style_map()` returns `(live, lost)`. The remove-before-rebuild ordering itself is pre-existing and affects the email and pptx paths too, so it is filed as **#192** rather than widening this PR.

## What the page has to get right

- **Opens on the mapping in force**, not empty. An empty form reads as "nothing is set" on a server whose master YAML has set a mapping all along, and the first save would silently clear it.
- **A configured style the base Word template does not define is still offered**, marked *not in the base Word template* — dropping it would lose configuration by rendering a page. A note beside the grid says the marker means "missing **here**": a Word template tool renders onto its own document, which may well define the style. Shown only when something is marked.
- **A master key the override drops is named** ("Not in force"), because the override replaces the master's mapping — otherwise an admin reading `docx_templates.yaml` on the volume sees a setting that simply is not happening.
- **Keys the renderer does not act on are flagged**, read from the master too, and kept *out* of "Not in force": they were never in force, and listing them beside a real one sends someone looking for a setting they never had.

`/styles/save` also fits `/{kind}/save`, so the routes are registered before the generic ones — as `/base` and `/files` are — and a test pins the ordering.

## What I decided against

- **Merging key-wise.** Above.
- **Preventing** the lost-tool case rather than reporting it. It is pre-existing, it affects three packages, and an ordinary template save still hits it — #192.
- **Locking the master file** against the TOCTOU that "saving writes every field" buys (a volume edit landing between the `GET` and the `POST` is overwritten). The page says to reload first; a lock over a file an admin may have open in an editor costs more than the window is worth. Recorded in the docs.
- **A free-text style field.** The dropdown is the base Word template's styles because that is what the static tool renders onto. The escape hatch is that an existing value outside that list is offered anyway.
- **Making the page per-kind-generic.** The storage is (all three kinds get a `_global.yaml` slot); the page is Word-only, because Word is the only kind with a top-level setting today.

## Verification

`ruff` clean, **2382 passed** (32 new). **23 mutations, all caught** — including two that initially *survived* and exposed a pair of vacuous tests of my own, which is the part worth reading:

- *"a save applies even when the files look unchanged"* froze the fingerprint **after** the first read, making the read following the save a cache *miss* rather than the hit it was meant to be. Deleting the explicit invalidation left it green.
- the live-reload test only ever **created and deleted** the override file, which moves its *existence*. A fingerprint blind to an edit but not to existence survived. There is now a test that rewrites the same file to the same length.

The load-bearing test is `test_a_saved_mapping_reaches_a_generated_document`: it asserts the paragraph style of a real `Document`, not that a YAML key moved.

`test_admin_assets.py`'s sweep gains `/styles` and `/files` — it claims to cover every rendered page and `/files` was already missing.

## Docs

`development/dynamic-templates` (new *Kind-wide settings* section), `tools/word` (which consumer resolves the map when, the measurement, why the fingerprint alone is not enough), `development/admin-ui` (the limitation removed, the page's rules, the TOCTOU), `shared-modules`, `architecture`, `templates.md`, the user `admin-ui` page, and `AGENTS.md`.


dvejsada:
@claude review

Areas I'd most want checked:

1. **Is "replace, not merge" right?** The whole design rests on it. The alternative — a key-wise merge, with the page showing the effective value and a "clear" that writes an explicit tombstone — was never seriously tried. Is there a case where an admin would reasonably expect the master's other keys to survive their save, and does the pre-filled form genuinely prevent the loss, or only usually?

2. **Un-caching `load_global_style_map()`.** It now does two `is_file()` checks plus up to two YAML parses per Word document build. I claimed that's negligible against document generation and once-per-build. Is it actually once per build on every path — does anything call it in a loop, or per paragraph, that I missed?

3. **The re-registration.** `resync_docx_style_map()` calls `register_docx_template_tools_from_yaml`, which re-registers every enabled template. Is there state that does *not* survive a re-registration — anything set on a tool after registration that this silently resets? And what happens to a template that fails to register: it's caught and logged inside the loop, but the page reports a count that says nothing about it.

4. **The reserved filename.** `read_spec_dir()` skips `_global.yaml` and `validate_name()` refuses `_global`. Are those the only two doors? `clone_spec`, `rename_spec`, `adopt`, the draft route — does any of them reach a spec path without going through `validate_name()`?

5. **Anything the page still lies about.** #161 was a lie-in-the-UI bug. I removed one and tried hard not to add another — but the dropdowns list the *base Word template's* styles, while the mapping also applies to documents rendered from a *per-template* docx. Is "not in the base template" the right marker, or does it now imply something false for a key only a template's own document defines?

---

dvejsada:
Answered two of my own questions while the review was running. Pushed `669fed1`.

### Q2 — un-caching `load_global_style_map()`: I was wrong

I claimed the cost was negligible. Measured it instead:

| | |
|---|---|
| Calls per document build | **1** (151-paragraph document, confirmed) |
| Full resolution | **~8 ms** |
| Markdown-parse step for that document | ~50 ms |

So it was ~15% of the render, not "negligible" — the master file is a few hundred lines of worked examples and parsing it is not free.

It now caches against the two config files' **mtime and size** (~0.02 ms on a hit, a 335× difference), so a hand edit on the volume is still picked up on the next document.

My original objection to a fingerprint still stands, so it's answered rather than dropped: two writes inside one filesystem timestamp tick that leave the file the same length look identical — swapping one six-letter style name for another does exactly that. So the *writer* says so. `invalidate_global_style_map()` is called from `resync_docx_style_map()`, beside the re-registration it already did. The fingerprint is for edits made outside the server; a write the server made itself, it knows about.

**Two of the tests I wrote for this were vacuous, and the mutations caught them.** Worth recording because both failed the same way — testing the mechanism rather than the failure:

- *"a save applies even when the files look unchanged"* froze the fingerprint **after** the first read, so the read following the save was a cache *miss*, not the hit it was supposed to be. Deleting the explicit invalidation left it green. Now frozen before the first read.
- the live-reload test only ever **created and deleted** the override file, which moves its *existence*. A fingerprint blind to an edit but not to existence survived. There is now a test that rewrites the same file to the same length, so only mtime distinguishes them.

Both mutations now fail. Updated the `tools/word` table and `AGENTS.md`, which had said "not cached".

### Q4 — the reserved filename: covered, and now pinned

Every store operation that names a spec builds its path through `_spec_path` → `validate_name`, so all six refuse `_global`:

```
save_spec    -> refused    set_enabled  -> refused
rename_spec  -> refused    get_spec     -> refused
clone_spec   -> refused    delete_spec  -> refused
```

with the settings file and the existing specs untouched. `adopt()` and the draft route reach the disk through `save_spec`, so they inherit it. That's now `test_no_store_operation_can_reach_the_reserved_file` — one guard, but it has to cover every door, and the test fails the day one of them builds a path itself.

**Q1, Q3 and Q5 are still open** and are the ones I'd most value a second opinion on — particularly Q3 (what a re-registration silently resets, and the failure a template hitting an exception in that loop reports to nobody).

---

dvejsada:
Review addressed. Pushed `197b08e` and `9818bae`.

### Q3 — the one actionable finding. Confirmed, and worse than the review says

I'd started digging into this independently while the review ran, and we converged on the same place. The review has it right that `_register_single_template` fails *gracefully* and the `False` is discarded. The part worth adding: `register_docx_template` calls `safe_remove_tool` **before** attempting the rebuild, so the failure is not "the tool stayed stale" — **the tool is gone**. Reproduced:

```
BEFORE: ['letter']
AFTER : []
PAGE SAYS: Saved. 1 key(s) overridden; 0 Word template tool(s)
           re-registered on the new mapping.
```

A registered template whose `.docx` was deleted on the volume since startup, and the admin was editing a *style*. Silently off the server, with a success message carrying a smaller number.

`resync_docx_style_map()` now returns `(live, lost)` from a before/after comparison of the live names, and the route words the outcome for what happened — naming the template, the likely cause and where to fix it, as a **warning**, not a success with a smaller count in it. Both directions mutation-tested (report nothing / report it as a success).

I deliberately did **not** fix the remove-before-rebuild ordering here. It is pre-existing, it affects the email and pptx paths the same way, and an ordinary template save through `AdminContext.register()` still loses the tool silently — so it wants its own change, not a widening of this one. Filed as **#192** with the reproduction and a build-then-swap suggestion.

### Q5 — taking this one, not as a nit

> maybe worth a one-clause addition […] but not blocking

Agreed on the reasoning, disagreed on the severity: "not in the base template" reading as "missing everywhere" when it means "missing here" is the exact shape of #161, reintroduced one layer down. Fixed rather than noted.

A note beside the grid now names the marked styles and says a Word template that defines the style in its own document still uses it — and that it is only a mistake where neither document has it. Shown only when something is actually marked; a note that is always there is a note nobody reads. Mutation-tested in both directions (dropped, and shown unconditionally).

### Q1 — TOCTOU recorded, not fixed

Right, and I'd rather it be written down than closed over. Locking a hand-edited file an admin may have open in an editor costs more than the window is worth. The page now says to reload before saving where a volume edit is possible, and `development/admin-ui.md` records that "saving writes every field" is what buys the genuine no-op, with the TOCTOU as its price.

The second gap you name — an unrecognised master key having no form field and dropping out on save — is intended, and is what `_unrecognised()` exists to announce before it happens. It reads the master as well as the effective mapping, precisely so an override cannot hide one.

### Q4 — the master-YAML `_global` wrinkle

Confirmed and agreed it's cosmetic. Such a template registers and runs, and can never be adopted (blocked at `get_spec` → `_spec_path` → `validate_name`). Worth being explicit that it is *harmless*, which is why I'm leaving it: it never writes to the `.d` layer, so the settings file is not at risk — the failure is an error page on Adopt, which is the correct outcome.

### Q2

Already answered in the comment above: I was wrong about the cost, measured it, and it now caches against an mtime+size fingerprint with explicit invalidation on the write path.

---

**2382 passing**, `ruff` clean. 23 mutations across the three pushes, all caught — including the two that initially *survived* and exposed a pair of vacuous tests of my own.

---

PR #77, opened 2026-06-27
feat(admin): FastHTML template-admin UI for docx + email templates

dvejsada:
## Overview

Adds an opt-in **FastHTML admin UI** for creating and managing dynamic **Word (docx)** and **email** templates from the browser — no YAML editing, no restart — aimed at non-technical users.

Everything is gated by `ADMIN_ENABLED`. **With the flag off (the default), runtime behaviour is unchanged**: `main.py` keeps the original `mcp.run()` path, no admin routes are mounted (`/admin` → 404), and no new code runs.

### Design decisions (confirmed with the maintainer)
- **Word authoring**: upload a real `.docx`; the UI auto-detects placeholders/conditionals — full Word fidelity, lowest risk.
- **Live updates**: admin runs **in the same uvicorn process** as FastMCP; saving a template (re)registers its MCP tool immediately via `local_provider` — no restart.
- **Access**: a single shared password (`ADMIN_PASSWORD`, falls back to `API_KEY`), signed session, constant-time compare.
- **Scope**: docx **+** email templates.
- **Topology/persistence** (defaults, open to change): single instance with an **abstracted storage layer**; per-template files under `config/<kind>_templates.d/` so the documented master YAML stays pristine.

## How it works (user flow)
1. Enable `ADMIN_ENABLED=true` + `ADMIN_PASSWORD`, browse to `http://localhost:8958/admin`, log in.
2. **Upload** a `.docx`/`.html` containing `{{placeholders}}` (and optional `{{#if flag}}…{{/if}}`).
3. The UI **scans the file** and auto-detects every placeholder/conditional, flags missing styles / orphan args / non-bool conditions, and **pre-builds the argument form**.
4. **Preview** with sample values (generates a real file in-memory; never uploaded anywhere).
5. **Save** → the template becomes a live MCP tool instantly.
6. **Edit** later, or **Replace document** to upload a new version and re-scan (existing args kept).
7. **Status** page for live tool counts, per-template usage, and a recent activity/error log.

## UI / UX
Built for non-technical users with a **self-contained theme** (no CDN — FastHTML's default pico.css is CDN-loaded and won't render offline/locked-down):
- Header bar + card layout, consistent page shell across login / list / upload / edit / status.
- **Edit page**: a "What we found" report with **chips** for detected placeholders and conditionals (conditionals highlighted + `if` badge); grouped *Tool details / Arguments / Advanced* cards; **Save / Preview / Cancel** bar; **Replace document** card.
[...]

PR #193, opened 2026-09-20
admin: make GET /admin reach the UI instead of 404ing

dvejsada:
Reported from a running beta. The server logs `[admin] Template-admin UI enabled at /admin` on startup, `docs/admin-ui.md` tells people to open `http://localhost:8958/admin`, and that exact URL came back **404**.

```
/admin        -> 404
/admin/       -> 303 /admin/login
```

Five commits: the fix, the guard against the *class* of bug, and three corrections — two from review, one from measuring a review claim I couldn't reproduce.

## The bug

`build_combined_app()` mounts two apps: `Mount("/admin", admin_app)` and `Mount("/", mcp_app)`. Starlette compiles the first to `^/admin/(?P<path>.*)$`, which **does not match the bare `/admin`**. Its router does have a missing-trailing-slash redirect, but it only fires when *nothing* matched — and the catch-all matches everything, so the request went to the MCP app, which correctly 404'd.

Isolated rather than inferred, one variable at a time:

| | Routes | `GET /admin` |
|---|---|---|
| **A** | admin mount alone | **307** → `/admin/` |
| **B** | A + `Mount("/", mcp_app)` | **404** |
| **C** | B + `redirect_slashes = False` | **404** |

A→B changes one thing and flips the result: the catch-all is the sole cause. C proves `redirect_slashes` was never reached — turning it off changes nothing.

## The fix

An explicit `Route` for the exact mount path, registered **before** both mounts, redirecting with 307 — the status and behaviour Starlette generates natively in case A. It reinstates the framework's own semantics that the catch-all suppressed rather than inventing anything, which is the argument for it being a fix and not a plaster.

Two details are load-bearing, both discovered by testing rather than reasoning:

- **The target comes from the request, not from `config.admin.path`.** *(Review finding — a real bug in my first version.)* Starlette builds its `Location` from the request scope, which folds in `root_path`. Mine built it from a build-time string: behind `uvicorn --root-path /office`, `GET /office/admin` answered `/admin/` — a path that doesn't exist on that deployment. Now verified identical to Starlette's output at `""`, `/office` and `/a/b`, and the test asserts against **Starlette's own output** rather than a literal so the two can't drift. It also preserves the query string.
- **`methods=["GET", "HEAD"]` is an invariant, not a default.** `ADMIN_PATH=/mcp` is legal, and MCP survives it *only* because the redirect doesn't take POST. Widened to all methods, an MCP client's POST gets a 307 and never gets a session — verified. A comment wouldn't hold that; a test does.

## The regression the guard prevents

The review suggested rejecting `ADMIN_PATH` values that collide with reserved paths, as hardening. Measuring it showed it isn't hardening — **without it this PR ships a regression**, and the review's stated reason was wrong:

| `ADMIN_PATH` | `POST /mcp` | `GET /healthz` | `GET /livez` |
|---|---|---|---|
| `/admin` | 200 +session | 200 | 200 |
| `/mcp` | **200 +session** | 200 | 200 |
| `/healthz` | 200 +session | **307** ← | 200 |
| `/livez` | 200 +session | 200 | **307** ← |

MCP is *not* the casualty. The probes are — and that 307 is mine: before this PR `Mount("/healthz")` didn't match the bare path so it reached the health route; `Route("/healthz")` does.

`config.RESERVED_PATHS` now blocks `/mcp` and the three probes, logging an error and falling back to `/admin`. **Falling back rather than refusing to boot**: the admin UI is optional and the MCP server isn't, so misconfiguring the former must not take document generation down — the same direction `is_enabled()` takes, logged for the same reason. Matching is exact (`/mcp2` is nobody's endpoint), and a test compares the list against the routes `main.py` actually registers so the copy can't drift.

## Why this survived ~10 merged PRs of admin-UI work

At `16d3616`, across the whole suite:

| Requests to | Count |
|---|---|
| `/admin/` | 16 |
| `/admin` | **0** |

Not an oversight repeated sixteen times — **one assumption counted twice**. In-app links come from `ctx.u("/")`, which already carries the slash; the tests were written from the route table with the mount prefix glued on in that same shape. The tests agreed with the app because both were built from the same model of the mount. Self-consistent, never externally checked. The one URL nothing in the system generates is the one a human types — and the one the startup log prints.

`test_admin_assets.py::test_every_page_is_rendered` is the sharpest illustration: it exists to sweep *every* admin page and asserts an exact count — and its index entry is `("index", "/admin/")`. Exhaustive, exhaustively inside the boundary.

## `tests/test_entry_points.py` — guarding the class

Two tests for `/admin` don't stop the next one. Every URL this server advertises — the three probes, `/mcp`, `/admin` — is asserted to resolve **exactly as advertised**, no trailing slash added to make it pass.

The rule it encodes: **its assertions are not written from the route table.** That is what the tests and the app already agreed about; another test drawn from it reproduces the blind spot at a new indentation. Each entry is a URL a *stranger* uses.

It also caught the same blind spot in two `/mcp` tests missing `follow_redirects=False` — one pre-existing, **one I added in this PR**, in the file whose whole subject is that distinction. Both fixed; both still pass, which answers the review's open question: `/mcp` matches directly, it was never a silent redirect.

## Provenance

Pre-existing since **#77**, the original admin-UI PR. Not from the recent template-lifecycle work.

## Verification

`ruff` clean, **2407 passed** (25 new). **13 mutations, all caught**, including the two that matter most:

| Mutation | Result |
|---|---|
| Remove the redirect route (reintroduce the bug) | 2 failed |
| Register it *after* the catch-all | 2 failed |
| **Build the target from config** (the `root_path` bug) | 2 failed |
| **Widen the methods** (kills MCP under `ADMIN_PATH=/mcp`) | 1 failed |
| Drop the query string | 1 failed |
| Remove the reserved-path guard | 5 failed |
| Prefix-match reserved paths (cry wolf on `/mcp2`) | 2 failed |
| Let `RESERVED_PATHS` drift from `main.py` | 1 failed |
| Redirect into open air / shadow `/mcp` / hard-code `/admin` | 1–6 failed |

One order-dependent failure found and fixed en route: a test entering a `TestClient` on the module fixture's app re-ran the lifespan and restarted the MCP session manager under every later test in the file. Noting it because "passes alone, fails in the suite" is the failure most likely to be waved off as flake.

## Docs

`development/admin-ui.md` gains a **Mounting** section: why the route order is load-bearing, why the target comes from the request, why the methods list is what it is, and what `test_entry_points.py` is for. `docs/configuration.md` gains the `ADMIN_PATH` constraint. `docs/admin-ui.md` needed no change — the URL it documents is the one this fixes.


dvejsada:
@claude review

The one-line fix is the easy part. What I most want challenged is the blind-spot analysis, and whether the fix is actually complete:

1. **Are there other unreachable entry points of the same shape?** The bug is "a URL the app never generates for itself, so nothing tests it." `/admin` was one. Check the rest of the surface for the same class: does `Mount("/", app=mcp_app)` swallow anything else it shouldn't — `/mcp` vs `/mcp/`, the `/healthz` routes, a bare `/admin/base` or `/admin/files`? I only fixed the one I was handed.

2. **Is 307 right, and is GET/HEAD right?** Starlette's `redirect_slashes` uses 307, so I matched it. But a `POST /admin` still falls through to the MCP mount and 404s. Is that acceptable, or should the redirect cover all methods so the behaviour is uniform? And does 307 interact badly with anything in front of this — a reverse proxy, a `root_path` deployment?

3. **`config.admin.path` normalisation.** `_normalize()` strips trailing slashes and forces a leading one, so `admin_root = config.admin.path + "/"` should always be well-formed. Is there an `ADMIN_PATH` value that breaks it — `/`, `//admin`, a path with a query or a space in it? What happens if someone sets `ADMIN_PATH=/mcp`?

4. **Did I actually diagnose it, or just make the symptom go away?** I claim the 404 came from the MCP app via the catch-all, not from Starlette failing to match. If the real cause is different — say FastMCP's own router, or `redirect_slashes` being disabled somewhere — then the redirect route is a plaster over something else and the next mount will hit it again.

5. **The blind-spot claim in the description.** I assert every test entered from inside the mount (16 requests to `/admin/`, 0 to `/admin`) because tests and in-app links were both built from the same model of the mount. Is that the real reason, or is there a test I have missed that *should* have covered this and failed to for a different reason? And is there a cheap structural guard better than my two tests — something that asserts the URL `docs/admin-ui.md` documents actually responds, the way `tests/test_config_docs.py` ties env vars to their docs?

---

dvejsada:
Answered Q1 and Q5 myself while the review runs. Pushed `11a535a`.

### Q1 — no other entry point has this shape

Probed the whole advertised surface against the **real** `main.mcp` (a fresh `FastMCP()` has no health routes, so my first probe 404'd on `/healthz` and told me nothing — worth stating, because that near-miss is the same trap as the bug):

| URL | bare | with slash |
|---|---|---|
| `/healthz` | 200 | 307 |
| `/readyz` | 200 | 307 |
| `/livez` | 200 | 307 |
| `/mcp` | 406 *(FastMCP on the `Accept` header — resolved)* | 307 |
| `/admin` | 307 → 200 | 303 → login |

`/admin` was the only one. Inner admin routes are fine in both forms — FastHTML's own router redirects `/admin/base/` → `/admin/base`.

### Q5 — there is a cheaper structural guard, and it is now in

You were right to push on this: two tests for the instance don't stop the next one. `tests/test_entry_points.py` guards the class.

The rule it encodes is that **its assertions are not written from the route table.** That is the whole point. The route table is what the tests and the app already agreed about — writing another test from it reproduces the blind spot at a new indentation. Each entry is instead a URL a *stranger* uses — a Kubernetes probe, an MCP client, a person typing the address the startup log prints — requested exactly as advertised, with no trailing slash added to make it pass, asserting only "not a 404" because what went wrong was never about the body.

It also covers the other side of the seam (the admin mount not shadowing `/mcp`), since the one line protecting one direction protects the other.

Mutations, all caught: reintroducing the 404; redirecting into open air (resolves, leads nowhere — passes the "not 404" test, fails the follow-through one); widening the admin mount so it shadows `/mcp`.

**2391 passing.** Q2, Q3 and Q4 are still open — Q4 especially, since if I mis-diagnosed the cause then this is a plaster and the next mount hits it again.

---

dvejsada:
Q2 and Q3 answered too. Pushed `0ee3d0d`.

### Q2 — GET/HEAD is load-bearing, not caution

I'd picked GET/HEAD as the conservative default and asked whether uniformity across all methods would be better. It would be actively wrong, and the reason only showed up when I tested Q3's `ADMIN_PATH=/mcp` case.

`ADMIN_PATH` is free text and nothing rejects `/mcp`. The MCP endpoint survives that config **only** because three things line up:

1. the redirect `Route` does not take POST,
2. `Mount("/mcp")` doesn't match the bare `/mcp`,
3. the catch-all still does.

So an MCP client's POST falls past both admin routes to the MCP app. Verified the counterfactual rather than reasoning about it — with the `Route` widened to all methods:

```
POST /mcp -> 307 /mcp/ | mcp-session-id: False
```

The client never initialises. That makes the methods list an **invariant, not a default**, and a comment wouldn't hold it — so there's now a test, and widening the list fails it. Someone tidying it for symmetry gets a red test instead of a silently dead endpoint.

A plain `POST /admin` still falls through to the MCP mount and 404s. That's correct and stays: there is no POST endpoint at the admin root, and a 404 from the wrong app is a far smaller problem than the above.

### Q3 — normalisation edge cases are pre-existing and none get worse

`AdminSettings._normalize()` handles the ordinary ones (`admin`, `/admin/`, `/`, `//`, empty, whitespace all → `/admin`). Three produce malformed values:

| `ADMIN_PATH` | normalised | redirect target |
|---|---|---|
| `///admin///` | `///admin` | `///admin/` |
| `/a b` | `/a b` | `/a b/` |
| `/admin?x=1` | `/admin?x=1` | `/admin?x=1/` |

All pre-existing, and in each case the `Route` and the `Mount` are built from the *same* value, so they agree — the redirect makes none of them worse, and the UI is equally unreachable with or without this PR. Tightening `_normalize()` is a separate change and I've left it out rather than widen this one; say if you'd rather it came in here.

### Still open

**Q4** is the one I most want a second pair of eyes on — whether I actually diagnosed the cause or just made the symptom go away. If the real mechanism is something other than the catch-all swallowing the request, this redirect is a plaster and the next mount hits the same thing.

**2392 passing**, 7 mutations caught across the three commits.

---

dvejsada:
Q4 answered — the one I was least willing to take on trust. Isolated it rather than reasoned about it, changing **one variable at a time**:

| | Routes | `GET /admin` |
|---|---|---|
| **A** | admin mount alone | **307** → `/admin/` |
| **B** | A + `Mount("/", mcp_app)` | **404** (body: `Not Found`) |
| **C** | B + `redirect_slashes = False` | **404** |

Three things follow, and together they close the question:

1. **Starlette's native behaviour is correct and was present all along.** In A, with nothing else registered, `redirect_slashes` fires and produces exactly 307 → `/admin/`.
2. **The catch-all is the sole cause.** A → B changes one thing: adding `Mount("/", mcp_app)`. That alone turns 307 into 404. Not FastMCP's internal router, not a config flag — the request matches the catch-all, so the router never reaches the no-match branch where `redirect_slashes` lives.
3. **`redirect_slashes` was never the lever.** Turning it off in C changes nothing, which is only possible if it was never consulted in B. Had it been implicated, C would differ from B.

### Why this makes it a fix rather than a plaster

The strongest evidence isn't the diagnosis, it's what the fix produces: **307 to `/admin/`** — the same status and the same target Starlette generates natively in case A. The `Route` doesn't invent behaviour or paper over a mystery; it reinstates the framework's own semantics that the catch-all suppressed. If it were compensating for something misdiagnosed, there'd be no reason for it to land on exactly the response the framework gives when the interference is removed.

It also means the general risk you were pointing at is real but bounded: **any** future mount under the catch-all inherits this, and none of them will get Starlette's redirect for free. That's now written down in the Mounting section, and `tests/test_entry_points.py` is the thing that catches it from outside rather than relying on someone remembering.

I considered the alternative fix this suggests — narrowing `Mount("/", mcp_app)` so it stops being a catch-all — and rejected it: the health routes are registered on the MCP app via `@mcp.custom_route` and need the root mount, so narrowing it is a far larger change with more ways to go wrong than the thing it fixes.

All five questions now answered. **2392 passing**, 7 mutations caught. CI running on `0ee3d0d`; I'll hold for the review's own read before merging, since Q4 is exactly where a second opinion is worth most.

---

dvejsada:
Both findings were right and both are fixed in `b065d2d`. The `root_path` one in particular is a genuinely good catch — it found a real bug in my fix *and* broke the argument I'd made for the fix being sound.

### `root_path` — confirmed, and it invalidated my Q4 conclusion

Reproduced rather than accepted:

| `root_path` | Starlette native | mine (before) |
|---|---|---|
| *(none)* | `/admin/` | `/admin/` ✓ |
| `/office` | `/office/admin/` | `/admin/` ✗ |
| `/a/b` | `/a/b/admin/` | `/admin/` ✗ |

Invisible in development, wrong the moment a prefix is stripped in front of the app — the browser is sent somewhere that doesn't exist on that deployment.

Worth being explicit about the damage to my own reasoning: I argued in the Q4 comment that this is a fix rather than a plaster **because it reinstates Starlette's native semantics**. Under `root_path` it didn't. The argument was right; my code didn't meet it.

It does now — `_admin_root_redirect` derives the target from `request.url`, matching native output at `""`, `/office` and `/a/b`. And the test asserts against **Starlette's own output rather than a literal**, so the invariant is "whatever the framework would have done if the catch-all weren't suppressing it", not a particular string that can drift.

Deriving from the request also preserves the query string, so a bookmarked `/admin?next=…` keeps its parameters. Pinned as well.

### `follow_redirects` — right, and it applies to a test *I added in this PR*

You flagged the pre-existing `test_mcp_endpoint_still_works`. The same flaw was in `test_the_mcp_endpoint_is_not_shadowed_by_the_admin_mount`, which I wrote **in this PR, in the file whose entire subject is this blind spot**. A test written to guard "can't tell a direct match from a followed redirect" that couldn't tell a direct match from a followed redirect.

Both now pass `follow_redirects=False`, and both still pass — which settles your open question: **`/mcp` matches directly**, it was never a silent redirect. (You flagged you couldn't execute code to check; that's the answer.)

### One more, found while writing the proxy test

It initially failed *only in file order*. Entering a `TestClient` on the module fixture's app ran the lifespan a second time and restarted the MCP session manager underneath every later test in the file. It builds its own app now. Noting it because "passes alone, fails in the suite" is the failure mode most likely to get waved off as flake.

### On your sandbox note

Worth saying: you found the highest-value defect here **without** being able to execute anything, by reading Starlette's documented semantics against the diff. Both findings were reachable only that way — I had a passing test suite and it told me nothing.

**2396 passing**, ruff clean. 9 mutations caught overall; the two new ones are building the target from config (fails the proxy and query tests) and dropping the query string.

---

dvejsada:
Q3's hardening idea taken, in `269000d` — but measuring it first changed what it is. **Without it, this PR ships a regression.**

### Correcting the premise

> every request to `/mcp` … is silently captured by the admin app instead. That's a full, silent outage of the MCP server

Not so — measured, not reasoned:

| `ADMIN_PATH` | `POST /mcp` | `GET /healthz` | `GET /livez` |
|---|---|---|---|
| `/admin` | 200 +session | 200 | 200 |
| `/mcp` | **200 +session** | 200 | 200 |
| `/healthz` | 200 +session | **307** ← | 200 |
| `/livez` | 200 +session | 200 | **307** ← |

MCP survives `ADMIN_PATH=/mcp`: `Mount("/mcp")` doesn't match the bare `/mcp`, and the redirect `Route` is GET/HEAD-only, so a client's POST falls through to the catch-all. (That is the three-way coincidence I described under Q2, and it holds.)

You were right that there's a problem — it's just in the probes, not MCP. And **the 307 there is mine.** Before this PR, `Mount("/healthz")` didn't match the bare path, so it fell through to the health route and returned 200. `Route("/healthz")` *does* match. So the redirect made that collision worse than it found it. That moves the guard from "nice hardening" to "in scope", which is why it's in this PR rather than a ticket.

### What landed

`AdminSettings._normalize()` rejects a path in `config.RESERVED_PATHS`, logs an error naming the collision, and falls back to `/admin`.

**Falling back rather than refusing to start** — the admin UI is optional, the MCP server isn't, so misconfiguring the former must not take document generation down with it. Same direction `is_enabled()` takes for an unrecognised value, and logged for the same reason.

The list is a copy (config stays free of application imports), so there's a test comparing it against the routes `main.py` actually registers — the `test_admin_style_keys.py` pattern. Matching is exact: `/mcp2` and `/healthz-ui` are nobody's endpoint and are left alone, mutation-tested in that direction too, since a guard that cries wolf would be worse than none.

Mutations, all caught: removing the guard; relocating silently; prefix-matching instead of exact; letting the list drift from `main.py`.

### On Q4 and Q5

**Q4** — you gave a static confirmation and were explicit that you couldn't reproduce. I ran the isolation afterwards and it matches your reading exactly: admin mount alone → 307; add the catch-all → 404; `redirect_slashes=False` on top → still 404, proving it was never consulted. Your conclusion was right without being able to execute it.

**Q5** — agreed the doc-parsing guard's value is the *other* failure mode (docs drifting to a URL nothing serves), which `test_the_bare_mount_path_reaches_the_admin_ui` doesn't cover. Not adding it here; this PR has grown enough. Worth a ticket if you want one.

**2407 passing**, ruff clean, 13 mutations caught across five commits.

---

PR #194, opened 2026-09-20
admin: organise the UI by product area, with tabs

dvejsada:
## What this is

The admin UI was arranged **by function**: one page listing every template of every kind, one page holding all five base templates, one for the Word style mapping, one for source files, one for status. So everything about Word was spread across three pages, PowerPoint across two, and Excel appeared only as one card buried in the middle of the base-templates page. The XML tool had no presence at all, and the top bar carried nine flat links with no grouping or active state.

It is now arranged **by what an admin is working on**.

| Section | Tabs |
|---------|------|
| Word `/word` | Overview · Templates · Base template · Style mapping |
| PowerPoint `/powerpoint` | Overview · Templates · Base designs |
| Excel `/excel` | Overview · Named styles |
| Email `/email` | Overview · Templates · Base wrapper |
| XML `/xml` | Overview |
| Server `/server` | Status · Source files · Activity log |

`/` becomes a dashboard of six tiles carrying live counts, calls and errors.

## How it is built

`admin/sections.py` is a join over the two tables that already existed — `kinds.py` (dynamic template kinds) and `base_templates.py` (fixed base slots). The nav, the tab bars, the routes and the dashboard all derive from it, so **adding a section is one row**.

Each former page became a **panel**: a list of cards with no shell, wrapped by `views.sections.section_page()`. `app.py`'s `_panel()` is the single (section, tab) → renderer map, and every route that re-renders a tab after a POST goes back through it — so an upload, a revert and a style save cannot drift into looking different from a plain GET of the same tab. A page reached *from* a tab but not itself a tab (an editor, a delete confirmation) stays a full page and names its section, so the nav keeps the right item lit.

**The Overview tab is why Excel and XML have pages at all.** Neither has a template to manage, but both expose a tool — and a tool that no page in the admin UI mentions is a tool nobody can check on. It reports the section's tools, whether each is live, and its calls/errors/warnings from `metrics`.

## The part worth reviewing carefully

A section slug shares the first path segment with a template kind, and `email` is deliberately **both**. Two rules keep a section route from eating a handler:

1. Section routes are registered first and are **GET-only**; every two-segment `/{kind}/…` route is a POST. That is what lets `GET /email` be the Email section while `POST /email/save` still reaches the save handler.
2. The slugs are **literal**, never a `/{slug}` pattern — a catch-all also matches `/new/docx`, and since `rt()` defaults to GET *and* POST it would have swallowed every save handler on the server. (It did, in the first draft.)

`sections.assert_slugs_free()` fails at startup on a slug that shadows a page. `tests/test_admin_sections.py` walks the **real route table** and fails if anything registered after the section routes is shadowed by one — the invariant itself, not the two rules that currently imply it.

Also documented where it bites: FastHTML fills *every* handler parameter from the request, so a captured value passed as a keyword default arrives as `None`. Both route registrations are factories for that reason.

## Visual

Token system extended from colour to spacing, type sizes, radii and shadows. The solid indigo slab is now a bordered sticky app bar with a brand mark and an active-pill; tabs are underlined links (bookmarkable, work with JavaScript off); cards put their own actions in a divided header strip; the status dot on a badge is reserved for *state*, so a classification beside it reads as a label. Light and dark both checked; at 500px there is no horizontal page scroll — the nav and tabs scroll, tables stay inside their cards.

## Compatibility

Every **action** route is byte-identical (`/{kind}/save`, `/{kind}/{name}/edit|clone|delete|reupload|preview`, `/base/{key}/upload|revert|download`, `/new/{kind}`), so editors, uploads and CSRF flows are untouched. `/status`, `/files`, `/styles` and `/base` redirect (303) to the tabs that replaced them — each was a top-bar link for the whole life of the UI, so they are in bookmarks and in pasted links. The test asserts the redirect *and* that its target renders.

## Testing

`ruff check .` clean. **2451 passed, 12 failed** — all 12 verified as pre-existing on `master` by stashing: 7 `test_pptx_text_metrics` (font metrics), 3 `test_admin_source_files` (Windows symlink privileges, `?` in a filename), 1 `test_admin_pptx` encoding, 1 `test_admin_log_view` capture-level.

New coverage in `tests/test_admin_sections.py`: every section and tab renders (parameterised over the table, so a new section is covered the moment it is added), the open tab and nav item are marked exactly once, a single-tab section draws no tab bar, unknown tabs redirect, every moved URL redirects to a page that renders, and the route-shadowing walk above.

Docs updated in the same change: `docs/admin-ui.md`, `docs/development/admin-ui.md`, `docs/templates.md`, and the **Admin UI** rules in `AGENTS.md`.


dvejsada:
@claude review

dvejsada:
Thanks — both points addressed in 3f14191.

**`status` missing from `RESERVED_SEGMENTS`:** you were right that it's only cosmetic today, but a hand-written list that's correct about three of four entries is the kind of asymmetry that invites someone to trust it and be wrong. So rather than adding the missing string, the legacy half is now **derived**: `MOVED_PATHS` moves to `admin/sections.py` (it's the same kind of fact as the table beside it — which tab a thing lives on), and `RESERVED_SEGMENTS` is the admin's own literal pages plus the first segment of every entry in it. Adding a redirect now reserves its segment on its own, and `test_every_moved_path_reserves_its_own_segment` pins that.

**`/base` landing on Word only:** correct reading — it is the accepted trade, not an oversight. It wasn't actually written down anywhere as a trade, though, which is the part worth fixing, so it's now stated in both `docs/admin-ui.md` (for the admin who follows an old bookmark) and `docs/development/admin-ui.md` (with the reasoning: the Word document is what someone arriving at `/base` is nearly always after, and the other four slots are one click away in their own sections).

**On the unverified test claim:** fair to flag. Re-run locally after this commit: `ruff check .` clean, **2452 passed, 12 failed**. The 12 are the same set I verified as pre-existing on `master` by stashing the branch — 7 `test_pptx_text_metrics` (font metrics on this box), 3 `test_admin_source_files` (Windows symlink privileges and a `?` in a filename), 1 `test_admin_pptx` encoding, 1 `test_admin_log_view` capture-level. CI's Lint & Test job is the independent check and it passed on the first commit.

dvejsada:
@claude review

PR #196, opened 2026-09-20
pptx: stop losing content on a customer template (#195)

### LINKED ISSUE #195: PowerPoint: a customer template loses content silently — columns dropped, text overflowing, drawn text unreadable
Building a deck on a corporate template (a Google Slides export whose design lives in full-bleed background pictures) produced a badly broken file. Investigating it turned up six separate defects, five of which fail **silently** — the caller gets a success response and no warning.

Reproduced against both the customer template and the two templates this repo ships.

### 1. `two_column` addresses placeholders by `idx`, and loses content when the numbering differs

`_build_two_column_slide` assumes PowerPoint's built-in numbering — 1/2 for Two Content, 1–4 for Comparison. That is a convention a template need not follow. The template in question numbered its two cards **4 and 2, left and right in that order**, and its comparison layout **4, 13, 14, 15**.

Result: the right column was written into the *left* card, the left column and both headings were dropped, three placeholders were left showing "Click to add text" — and **nothing was reported**. Both spellings were affected; the no-headings path lost a column too.

### 2. No warning on any of it

AGENTS.md is explicit: *"A branch that skips, substitutes or degrades something the caller asked for reports it… never the log alone."* This path dropped two headings and a whole column and returned zero warnings.

### 3. Autofit is measured against a hardcoded 18pt

`_fit_text` always passed `DEFAULT_BODY_FONT_SIZE` to `estimate_text_fill`. **Both shipped templates set 28pt** in the master's `<p:bodyStyle>`, so every estimate was low by the square of the ratio (~2.4×). Text needing 1.9× its placeholder measured as 0.86×, no `fontScale` was written, and PowerPoint renders a bare `<a:normAutofit/>` at full size until someone clicks into the box — so the text ran off the bottom of the slide. The overflow warning is derived from the same number, so it never fired either.

This affects every deck built on the shipped templates, not just the customer one.

### 4. Four content placeholders is not enough to mean "Comparison"

`classify_layout()` claimed `ROLE_COMPARISON` for any layout with four content placeholders. The customer template spends four on a layout that is **three cards side by side plus a caption bar**, so it took the comparison role and every two-column slide in the deck landed on three cards.

Related: when a role is unprovided the fallback is the positional index, which is a guess about a template that has already proved unusual. Position 4 there is Title Only — no body placeholder at all — so the slide lost both columns rather than merely looking wrong.

### 5. Unfilled placeholders are left behind

`add_slide()` copies every placeholder the layout defines. A layout offering more than the slide filled leaves empty "Click to add text" boxes — the third card, an unused heading strip, a Section Header's body (a `section` slide carries only a title), a title slide's absent subtitle. Three on one slide here. They do not print, but they are the first thing anyone opening the file to edit sees.

### 6. Drawn text and table fills ignore the template

Anything the builder *draws* rather than places in a placeholder is a plain text box, so it inherits `<p:defaultTextStyle>` — `tx1`, black — not the body style. On a dark template the KPI figures and timeline detail lines were **black on near-black**. Chart text is worse: it lives in its own part and inherits nothing, so scatter axis labels were black whatever the deck looked like.

Table colours were pinned to literals (`TABLE_HEADER_FILL = 4172C4`, Office's old default blue), so a table came out that blue on every template — including one whose own accent is a brand teal. `_set_cell_fill()` already wrote a theme name as `schemeClr`; only the default was not one.

### Also, separately

`closing` slides were hardcoded to `ROLE_TITLE`, so they always used the cover layout and there was no way to point them at a template's own contact layout short of naming it on every slide.

---

Fixes on `claude/friendly-ramanujan-vxejrj`.

dvejsada:
Fixes #195.

A deck built on a corporate template came out badly broken. Investigating it turned up six defects, five of which failed **silently** — the caller got a success response and no warning. Two of them affect the templates this repo ships, not just the customer one.

Every fix is verified against the real template, and the test for each one fails on the code before it.

### Two-column slides lost a column and both headings

`_build_two_column_slide` addressed placeholders by `idx`, assuming PowerPoint's built-in numbering. The customer template numbered its two cards **4 and 2 — left and right in that order** — and its comparison layout **4, 13, 14, 15**. The right column was written into the left card, the left column and both headings were dropped, three placeholders were left showing "Click to add text", and nothing was reported. The no-headings path lost a column too.

`content_columns()` reads the layout's shape instead: placeholders overlapping horizontally are one column, a column's tallest is its body, a shorter box above it is the heading strip.

### Nothing warned

Per the warnings rule in AGENTS.md, every shortfall now reports itself — `heading_inlined` (info: no heading strip, so it becomes a bold first line), `columns_merged` (warning: one content area, columns merged in order), `column_dropped` (error: nothing to write into). A test asserts the invariant directly: the content is in the file, or a warning says why not.

### Autofit measured against a hardcoded 18pt

`_fit_text` always passed `DEFAULT_BODY_FONT_SIZE`. **Both shipped templates set 28pt**, so every estimate was low by the square of the ratio (~2.4×). Text needing 1.9× its box measured as 0.86×, no `fontScale` was written, and PowerPoint renders a bare `<a:normAutofit/>` at full size until someone clicks into it — so the text ran off the slide. The overflow warning comes from the same number, so it never fired either.

`read_body_font_size()` resolves the size the way PowerPoint inherits it. `test_shipped_template_body_is_not_the_built_in_default` pins the 28pt so the assumption cannot drift back unnoticed.

### "Comparison" was decided by counting placeholders

Four content placeholders claimed `ROLE_COMPARISON`. The customer template spends four on **three cards plus a caption bar**, so every two-column slide landed on three cards. A layout now has to resolve to exactly two columns each with a heading strip; anything else returns `None` and is selectable by name only.

That change alone made things worse at first: with `comparison` unprovided the fallback was the positional index, and position 4 there is Title Only — no body placeholder at all — so the slide lost both columns rather than merely looking wrong. `ROLE_ALTERNATIVES` now tries a near-neighbour role the template actually has first (`comparison` → `two_column` → `content`), each of which can still hold the text.

### Empty placeholders were left in the file

`_drop_unused_placeholders()` sweeps them after the build — the third card, an unused heading strip, a Section Header's body, a title slide's absent subtitle. Placeholders holding a picture, table or chart are kept; footer chrome is left for the footer pass, which runs afterwards.

### Drawn text and table fills ignored the template

Text the builder draws is a plain text box, so it inherited `<p:defaultTextStyle>` (`tx1`, black) — on a dark template the KPI figures and timeline captions were black on near-black. Chart text was worse: it lives in its own part and inherits nothing. Both now take the master's body colour, kept as an `MSO_THEME_COLOR` so it keeps tracking the theme.

Table colours were literals (`TABLE_HEADER_FILL = 4172C4`), so a table came out Office blue on every template. They are now the theme names `accent1` and `bg2` — `_set_cell_fill()` already wrote a theme name as `schemeClr`, only the default was not one. Per-slide `header_color`/`fills` and the registry's `table` defaults still win, and a test pins that.

### Separately: `closing` had no role

It was hardcoded to `ROLE_TITLE`, so closing slides always used the cover layout with no way to redirect them per template. `ROLE_CLOSING` is the first entry in `CONFIGURED_ROLE_DEFAULT`: a role classification never assigns, supplied by the registry's `layouts:` mapping. Unmapped it resolves to `title` **without** a warning — that is its documented default, not a substitution. `missing_roles()` and the admin's analysis both exclude such roles, or every template would be reported as lacking something no placeholder arrangement can supply.

## Verification

- **2512 tests pass**, up from 2407; `ruff` clean. Four new test files, each written to fail against the code it fixes.
- All 14 slide types build on the corrected customer template with **zero warnings**, checked by rendering through LibreOffice and inspecting every slide.
- On the **original, uncorrected** template the same deck now keeps every word, with each compromise reported — that is the check that these are code fixes and not template workarounds.

## Notes for the reviewer

- **Behaviour change:** table headers move from a fixed blue to the template's `accent1`. On the shipped templates that is `156082` rather than `4172C4`, so existing decks will look different — deliberately, toward the template's own palette. `TABLE_HEADER_TEXT` stays an explicit white; choosing it from the theme needs a luminance decision the tool cannot make safely.
- **Docs are in the same commits**, per AGENTS.md. The `idx` assumption was written down as both an invariant and a known limitation in `docs/development/tools/powerpoint.md`; both are replaced. `AGENTS.md` gains two rules, and `docs/templates.md` and `docs/powerpoint-slides.md` cover the caller-facing parts.
- **History was rewritten once** before this PR opened: the commits originally cited `#194`, which is an unrelated admin PR merged before this branch existed. No one else had the branch.

https://claude.ai/code/session_01UiV7z58sAHrFbTr5VXF6ie

---
_Generated by [Claude Code](https://claude.ai/code/session_01UiV7z58sAHrFbTr5VXF6ie)_


dvejsada:
@claude review

Worth a close look at:

- `content_columns()` in `pptx_tools/placeholder_style.py` — the clustering thresholds (50% horizontal overlap for "same column", 60% height ratio for "heading strip"). Are there real template shapes these get wrong?
- `_drop_unused_placeholders()` — it removes any non-chrome placeholder with an empty text frame. I believe filled picture/table/chart placeholders are safe because they are no longer `<p:sp>` with a text frame, but that is the riskiest assumption in the PR.
- `ROLE_ALTERNATIVES` — whether any chain can put a slide somewhere that cannot hold its content.
- The table colour change is user-visible on existing decks (fixed blue → the template's `accent1`).


---

dvejsada:
Thanks — all four addressed in e8247b4. One turned out to be a real fix rather than a comment.

**Clustering against `column[0]` — fixed, with a test that fails on the old code.**

You were right that it's a live gap, but the self-limiting argument only covers the classification path. `classify_layout()` is indeed safe, because it demands exactly two columns with a heading each. But `_build_two_column_slide()` also calls `content_columns()` — on whatever layout the slide actually landed on, which may never have been vetted by `classify_layout()` at all (an explicit `layout:` override, or a layout resolved for `ROLE_TWO_COLUMN`). It then fills `columns[0]` and `columns[1]`. A spurious column there puts content in the wrong box, silently, which is the exact failure this PR exists to remove.

The reproducer is a column headed by a *narrow* strip: it pins the column's identity to the heading's width, so a box sitting inside the wider body below it starts a third column.

```
(1.0, 1.5, 2.0, 0.5)   narrow heading, spans 1.0–3.0
(1.0, 2.2, 5.0, 3.5)   its body, wider, spans 1.0–6.0
(3.5, 2.2, 2.5, 3.5)   inside the body's span, outside the heading's
(7.5, 2.2, 4.0, 3.5)   the genuine second column
```

Before: three columns (`[(1.0, 5.0), (3.5, 2.5), (7.5, 4.0)]`). After: two. Each column now carries a running span that widens as boxes join it. `test_a_column_is_matched_on_its_whole_extent_not_its_first_box` covers it.

**Heading is only looked for above the body** — documented in `content_columns()`, including what happens instead: a below-captioned layout reads as headingless, the captions become body text, and a slide that supplied its own headings gets `heading_inlined`. I left the behaviour alone deliberately — looking downward as well would risk reading a genuine caption as a heading, and the headingless path already keeps every word.

**`_drop_unused_placeholders()` docstring** — tightened. It now describes the actual test (`has_text_frame` plus empty), names the picture path as the only one that produces a graphic-holding placeholder today, notes that `_add_title_content_slide()` removes the placeholder outright for tables and charts, and says why a future insert-into-placeholder change would be safe for the same reason.

**`TABLE_ALT_ROW_FILL` luminance caveat** — added next to it, matching the note `TABLE_HEADER_TEXT` already carries.

On `ROLE_ALTERNATIVES`: agreed, and your reasoning matches mine — `title_style()` and `content_area()` resolve per *template* rather than per slide, so the chains cannot strand a title or a body. The genuine last resort (a layout with no body placeholder at all) is always reported.

2513 tests pass; `ruff` clean. Also re-verified against the real customer template: the corrected one still resolves every role and builds the original deck with zero warnings, and the uncorrected one still keeps every word with each compromise reported.


---

PR #197, opened 2026-09-20
admin: count master-YAML templates on Overview; pptx: say what a slide type takes

dvejsada:
Two fixes, both found by reviewing the **deployed** instance after #194 shipped. They are unrelated in code but came from the same look, so they are one PR with a commit each.

---

## 1. The Overview did not count hand-written templates

The deployment has the shape that exposes this: one Word template declared in `config/docx_templates.yaml`, none managed by the UI. Three pages then disagreed about the same fact:

| Page | Said |
|---|---|
| Word ▸ Templates | `smlouva_o_poskytovani_sluzeb` — **Live** — `master YAML` ✅ |
| Word ▸ Overview → "Tools the AI can call" | only `create_word_document` — **the live tool absent** ❌ |
| Dashboard → Word tile | **0 LIVE** ❌ |
| Server ▸ Status | **1 LIVE WORD TOOLS** ✅ |

`section_facts()` built its counts and its tool list from `store.list_specs()` — the managed half only. `template_table()` walks the managed specs *and* `unmanaged_master_specs()`, which is why the Templates tab was the one page that was right. A hand-written template registers the same MCP tool as one made here, so it belongs in both.

Both halves now feed the record. `ToolFact.origin` keeps them distinguishable — **From master YAML** rather than *From a template*, because one is editable here and the other is not, and knowing which saves a trip to the Templates tab to find out why a row has no Edit button. The note above the table used to say master templates are shown read-only elsewhere, which read as *"and are not counted here"*; it now says they are counted.

**On the tests:** they register the template the way `main.py` does at startup. A fixture that only writes the YAML has a template the registry has never heard of, so `live 0` is *correct* there and the test passes against the bug — which is what my first draft did, and why it's worth calling out.

## 2. A rejected slide field now says what that type does take

The same review turned up a real failed call on the server:

```
Invalid slides: slide 0 -> body: Extra inputs are not permitted;
slide 2 -> subtitle: Extra inputs are not permitted; slide 3 -> subtitle:
… slide 7 -> body: Extra inputs are not permitted
```

A seven-slide deck lost whole — the schema is strict, so one field on the wrong type fails the entire call and nothing is generated. The caller assumed what most people assume: that every slide takes a title, a subtitle and a body. It put `body` on the title slide and `subtitle` on the content slides.

**The published schema was not the gap.** It already names the types per field — `subtitle` renders as `[title] Subtitle: author, tagline, date… [closing] Line under the closing title.` The gap is that the error names the mistake and not the fix, and by the time it arrives the caller has read the schema once and drawn the wrong conclusion. Repeating the rule where it failed is what turns a retry into a correction:

```
slide 1 -> subtitle: 'subtitle' is not a field of a 'section' slide
(it takes: title, notes, layout); 'subtitle' belongs to: title, closing.
```

Both sides are derived from `_SLIDE_MODELS`, so a new slide type is covered without touching the error path, and a field no type declares says so rather than pointing nowhere. Only `extra_forbidden` is rewritten — every other error keeps pydantic's own wording, which already says something useful, and there is a test pinning that.

The tool description gains one paragraph on the same point, **including that the call fails whole** — worth knowing before writing eight slides — and `docs/powerpoint-slides.md` a matching line under its per-type field table.

Also corrects `coerce_slides()`'s docstring, which claimed MCP callers never reach this path because FastMCP validates first. They always reach it: the parameter is `Any` carrying a hand-built schema, so nothing validates earlier. These strings are the tool's error message, not developer diagnostics.

---

## Testing

`ruff check .` clean. **2459 passed, 12 failed** — the same pre-existing set verified on `master` by stashing (7 `test_pptx_text_metrics` font metrics, 3 `test_admin_source_files` Windows symlink/filename, 1 `test_admin_pptx` encoding, 1 `test_admin_log_view` capture-level).

8 new tests. The three admin ones fail without the fix and pass with it (checked by stashing only the two source files). The pptx ones pin the exact production payload.

Docs updated in the same change: `docs/development/admin-ui.md`, `docs/development/tools/powerpoint.md`, `docs/powerpoint-slides.md`, and the `create_powerpoint_presentation` description in `main.py`.


dvejsada:
@claude review

dvejsada:
Thanks — no changes needed for the two fixes themselves, but the 'latent fragility' is worth reporting back on, because it is not latent.

> the field/slide-type disambiguation relies on no field ever being named the same as a `SLIDE_TYPES` value (e.g. a future field literally called `image` or `table`). If that ever happens…

It already has. **`title` is both a field on every slide and a slide type.** The path was built by dropping every segment matching a type name, so the one field whose name collides was being silently removed from it:

```
>>> coerce_slides([{'type': 'content', 'title': {'bad': 1}, 'body': '- a'}])
Invalid slides: slide 0: Input should be a valid string
```

No field named — on the field every single slide has. Fixed in d8ceade: the discriminator tag is always at position 1 of a tagged union's path, so position identifies it and exactly one leading segment is dropped. A collision cannot recur however the field names change, which is stronger than the comment you suggested.

```
Invalid slides: slide 0 -> title: Input should be a valid string
```

Two tests added: one for a field named like a slide type keeping its name, one for a nested path (`series.0.values.0`) surviving the tag being dropped. Pre-existing rather than introduced here, but this PR builds on that filter, so it is this PR's to fix.

Re-ran locally: `ruff check .` clean, and the pptx/schema selection is **603 passed, 8 failed** — all 8 the pre-existing set (7 `test_pptx_text_metrics` font metrics, 1 `test_admin_pptx` encoding). Noted that you could not run the suite yourself; CI's `lint-and-test` is the independent check.

